PR #19, opened 2019-08-30
repl: reset tty cooked mode on ^D

bnoordhuis:
Before this commit, pressing ^D twice would exit the REPL but leave
the terminal in raw mode. Fix that by calling exit() directly instead
of closing the SIGINT listener.

That leaves just one reference to the `sigint_h` variable in repl.js,
the assignment in termInit(), but stepping through the code suggests
that's okay: the garbage collector marks it but doesn't collect it.

Fixes: https://github.com/saghul/quv/issues/18

### DIFF
--- a/src/repl.js
+++ b/src/repl.js
@@ -411,7 +411,7 @@ import * as uv from "uv";
     function control_d() {
         if (cmd.length == 0) {
             stdout.write("\n");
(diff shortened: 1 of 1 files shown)


saghul:
Cheers Ben!

PR #20, opened 2019-09-02
Update quickjs

saghul:


src/quv/modules.c
```diff
@@ -114,22 +115,81 @@ JSModuleDef *quv__load_http(JSContext *ctx, const char *url) {
...
+    CHECK_EQ(JS_VALUE_GET_TAG(func_val), JS_TAG_MODULE);
+    m = JS_VALUE_GET_PTR(func_val);
+
+    module_name_atom = JS_GetModuleName(ctx, m);
+    module_name = JS_AtomToCString(ctx, module_name_atom);
+    JS_FreeAtom(ctx, module_name_atom);
+    if (!module_name)
+        return -1;
+    if (!strchr(module_name, ':')) {
+        strcpy(buf, "file://");
```
bnoordhuis: Suggestion: maybe use `pstrcpy()` instead? When I see `strcpy()`, I always have to do a double take to check whether the buffer is big enough to hold the input. In this case it does but it's not _immediately_ obvious.

saghul: Good call. This one came straight from upstream, but yeah, makes sense!

saghul: https://github.com/saghul/quv/pull/21

PR #28, opened 2019-09-04
test: add initial tests

saghul:


tests/run.js
```diff
@@ -0,0 +1,6 @@
+import { run } from './t.js';
+
+import './test-args.js';
+import './test-noop.js';
```
bnoordhuis: Don't you want to run each test in a separate process?

The thing I despise about mocha the most is when a test clobbers some global state making another test fail. I've sunk a lot of time in tracking down such bugs in various test suites.

saghul: I did think of that, yeah. But I then thought "meh, I'll see how to cross that river if I get there :-)"

So for now let's see how it goes, nothing is cast in stone anyway!

PR #48, opened 2019-09-14
HTTP server

brandonros:
Part of https://github.com/saghul/quv/issues/44

I copied your `Udp` class to make a `HttpServer` class that takes `(host, port)` as arguments.

Now I'm working on blending https://github.com/parachvte/libuv-httpserver/blob/master/httpserver.c and https://github.com/h2o/picohttpparser#phr_parse_request

`HttpServer` will call `onrequest` when a request is parsed through `picohttpparser`


saghul:
Hey Brandon,

Before you make it any further: as I mentioned in https://github.com/saghul/quv/issues/44#issuecomment-531504791 I really don't want a homegrown HTTP server right now.

It has wasy too much complexity (if we are going to do it right) which I'd rather not have, and stand on shoulders of giants.

HTTP 1 (and 1.1) is reasonably simple, but then we also need HTTPS. And WebSockets, and HTTP/2. And in the meantime, HTTP/3 is on the way, using QUIC.

That's just too much. If we use something like libwebsockets, we get HTTP(S), WS and HTTP/2 for free. And thanks to the libuv integration we don't need to worry about threads or another event loop, pretty much like the integration with CURL.

Is this something that spikes your interest?

PR #74, opened 2019-10-19
fix example to be executable

andyjia:


andyjia:
thank you pointing it out. yes, I tried but failed. sorry I recall.

saghul:
No worries! I’ll try to go through the examples and fix then soon!

PR #84, opened 2019-11-07
Implement window.prompt and window.alert

### LINKED ISSUE #81: Add window.prompt
With an optional dependency on ```readline```.

TheBrokenRail:
```window.prompt``` currently has issues with the REPL, but otherwise works.

Fixes #81

### CODE REVIEW

File: src/misc.c [hunk 1]
```diff
@@ -333,7 +404,8 @@ static const JSCFunctionListEntry tjs_misc_funcs[] = {
     JS_CFUNC_DEF("tmpdir", 0, tjs_tmpdir),
     JS_CFUNC_DEF("exepath", 0, tjs_exepath),
     JS_CFUNC_MAGIC_DEF("print", 1, tjs_print, 0),
-    JS_CFUNC_MAGIC_DEF("printError", 1, tjs_print, 1),
+    JS_CFUNC_MAGIC_DEF("alert", 1, tjs_print, 1),
```
saghul: THis change breaks `console.warn`

CMakeLists.txt
```diff
@@ -65,6 +66,11 @@ if(NOT DISABLE_CURL)
     endif()
 endif()
 
+if(NOT DISABLE_READLINE)
+    include(FindReadline.cmake)
+    find_package(CURL)
```
saghul: this is incorrect

src/misc.c
TheBrokenRail: Fixed!

CMakeLists.txt
TheBrokenRail: Fixed when switching to REPLXX.

TheBrokenRail:
Switched to RERPLXX, which happens to work with the REPL better too!

src/misc.c
```diff
@@ -282,6 +285,54 @@ static JSValue tjs_print(JSContext *ctx, JSValueConst this_val, int argc, JSValu
     return JS_UNDEFINED;
 }
 
+#define STR_BUFFER_SIZE 128
+#define STR_BUFFER_INCREMENT 16
+
+static JSValue tjs_prompt(JSContext *ctx, JSValueConst this_val, int argc, JSValueConst *argv) {
```
saghul: I think it would better to just use 1 argument here and do the concatenation into a single string in JS. That would save us from all the realloc-ing.

TheBrokenRail: Turns out ```window.prompt``` is actually ```window.prompt([message[, defaultValue]])``` according to https://developer.mozilla.org/en-US/docs/Web/API/Window/prompt so I implemented that.

src/misc.c
```diff
@@ -334,6 +370,8 @@ static const JSCFunctionListEntry tjs_misc_funcs[] = {
     JS_CFUNC_DEF("exepath", 0, tjs_exepath),
     JS_CFUNC_MAGIC_DEF("print", 1, tjs_print, 0),
     JS_CFUNC_MAGIC_DEF("printError", 1, tjs_print, 1),
+    JS_CFUNC_MAGIC_DEF("alert", 1, tjs_print, 1),
+    JS_CFUNC_DEF("prompt", 0, tjs_prompt),
```
saghul: this needs to be 2, since it takes 2 arguments

src/misc.c
```diff
@@ -282,6 +285,39 @@ static JSValue tjs_print(JSContext *ctx, JSValueConst this_val, int argc, JSValu
...
+    replxx_end(replxx);
+
+    if (argc > 0) {
+        JS_FreeCString(ctx, message);
+    }
+    if (argc > 1) {
+        JS_FreeCString(ctx, default_value);
+    }
+
+    return JS_NewString(ctx, str);
```
saghul: this leaks `str`, since `JS_NewString` will create a copy.

src/misc.c
```diff
@@ -282,6 +285,39 @@ static JSValue tjs_print(JSContext *ctx, JSValueConst this_val, int argc, JSValu
     return JS_UNDEFINED;
 }
 
+static JSValue tjs_prompt(JSContext *ctx, JSValueConst this_val, int argc, JSValueConst *argv) {
+    char *str;
+
+    const char *message = "";
```
saghul: better initialize them to NULL

src/misc.c
```diff
@@ -282,6 +285,39 @@ static JSValue tjs_print(JSContext *ctx, JSValueConst this_val, int argc, JSValu
...
+    const char *default_value = "";
+
+    if (argc > 0) {
```
saghul: you can drop this check, QuickJS guarantees that it will fill the number of arguments you set when defining the function (see my last comment)

src/misc.c
```diff
@@ -282,6 +285,39 @@ static JSValue tjs_print(JSContext *ctx, JSValueConst this_val, int argc, JSValu
...
+    
+    if (argc > 1) {
+        default_value = JS_ToCString(ctx, argv[1]);
+        replxx_set_preload_buffer(replxx, default_value);
+    }
+
+    str = strdup(replxx_input(replxx, message));
+    replxx_end(replxx);
+
+    if (argc > 0) {
```
saghul: you can drop these chacks if you initialize the variables to NULL

src/misc.c
```diff
@@ -282,6 +285,39 @@ static JSValue tjs_print(JSContext *ctx, JSValueConst this_val, int argc, JSValu
...
+    }
+
+    Replxx* replxx = replxx_init();
+    
+    if (argc > 1) {
+        default_value = JS_ToCString(ctx, argv[1]);
+        replxx_set_preload_buffer(replxx, default_value);
+    }
+
+    str = strdup(replxx_input(replxx, message));
```
saghul: No need to dup the string, is there? `JS_NewString` will copy it anyway. Also, is anything else needed for freeing the return value of `replxx_input` ?

src/misc.c
TheBrokenRail: Both arguments are optional.

src/misc.c
TheBrokenRail: Both arguments are optional.

src/misc.c
TheBrokenRail: Both arguments are optional.

src/misc.c
TheBrokenRail: Fixed!

src/misc.c
TheBrokenRail: Fixed!

src/misc.c
TheBrokenRail: The ```message``` argument is optional.

TheBrokenRail:
When will this be merged?

CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ include(${CMAKE_SOURCE_DIR}/src/js/CMakeLists.txt)
 option(libuv_buildtests "" OFF)
 add_subdirectory(deps/libuv EXCLUDE_FROM_ALL)
 add_subdirectory(deps/quickjs EXCLUDE_FROM_ALL)
+add_subdirectory(deps/replxx EXCLUDE_FROM_ALL)
```
saghul: Set these options please, so we only build and link with the static library: https://github.com/AmokHuginnsson/replxx/blob/master/CMakeLists.txt#L7 Also diasble the tests.

TheBrokenRail: Done!

CMakeLists.txt
```diff
@@ -30,6 +30,8 @@ include(${CMAKE_SOURCE_DIR}/src/js/CMakeLists.txt)
 option(libuv_buildtests "" OFF)
 add_subdirectory(deps/libuv EXCLUDE_FROM_ALL)
 add_subdirectory(deps/quickjs EXCLUDE_FROM_ALL)
+option(REPLXX_BuildExamples "" OFF)
```
saghul: disable also building the dynamic library please, since we don't use it.

TheBrokenRail: Done!

PR #87, opened 2019-11-12
Support Cross Compiling

TheBrokenRail:
This allows you to specify a custom path to ```qjsc``` which is useful when cross-compiling and the version built by ```cmake``` won't run on the build system.


saghul:
This needs a default which uses the current build directory, otherwise the buid is broken,

TheBrokenRail:
Fixed!

PR #90, opened 2019-11-23
Add LICENSE File

TheBrokenRail:
While the license is stated in the source code, it would be nice to have a ```LICENSE``` file to quickly check without looking in the source code.


TheBrokenRail:
@saghul

saghul:
Oops, I had forgotten, cheers!

PR #94, opened 2019-12-26
Rename 'hrtime' to 'hrtimeBigInt'

stevefan1999-personal:
According to Node docs, hrtime should return a tuple of size 2 `[seconds: number, nanoseconds: number]`, and it is hrtimeBigInt to actually be the full size with nanoseconds.

### DIFF
--- a/src/js/performance.js
+++ b/src/js/performance.js
@@ -13,7 +13,7 @@ class Performance {
     }
 
     now() {
-        return tjs.hrtimeBigInt() - this._startTime;
+        return Number(tjs.hrtimeBigInt() - this._startTime);
     }
 
     mark(name) {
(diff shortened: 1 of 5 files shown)


saghul:
txiki.js doesn’t aim to be compatible with node. In addition BigInt is always supported here so there is no need to use a suboptimal solution IMHO.

cjihrig:
It looks like [Deno is working on a Node compatibility shim](https://github.com/denoland/deno/issues/2644), and the README in https://github.com/denolib/node states that it was merged into Deno. I don't follow Deno closely enough to know all the details, but if Node compatibility ever becomes something you want here, that might at least offer some inspiration.

saghul:
This makes sense @cjihrig, thanks for sharing! I'll reopen and take another look when I'm back up to speed.

saghul:
I gave this some more thought. In https://github.com/saghul/txiki.js/pull/105 I'm making performance.now() spec compliant so that should be ok now. As for Node compatibility, it should be easy enough to create a "node hrtime" based off the current one, given that BigDecimal is available now.

PR #96, opened 2019-12-26
Add fs to globals

stevefan1999-personal:
We should have some globals by default, rather than sticking them to tjs object. Right now I see fs can be one to be exposed for Node compatibility.


saghul:
I don’t really want to do this. fs does not aim to be Node compatible, so exposing it as a global would be misleading. Not polluting the global namespace and having a single tjs as a global is a design choice I’ll be sticking to, for now.

PR #97, opened 2019-12-26
Add some exotic examples

stevefan1999-personal:
So I said we could run some famous modules with Browserify as a shim to the require system and here they are, JSDOM and cheerio. 
I plan to add more examples in the future, but right now this is blocked because of the dependency on patch-hrtime branch to fix a bigint issue, so only if patch-hrtime can be passed we can discuss it a little bit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 BUILD_DIR=build
-BUILDTYPE?=Release
+BUILDTYPE?=MinSizeRel
 
 all: build
 
@@ -22,5 +22,9 @@ format:
 
 test:
 	./$(BUILD_DIR)/tjs tests/run.js
+
+test-advanced:
+	cd tests/advanced && npm install
+	./$(BUILD_DIR)/tjs tests/run-advanced.js
 
 .PHONY: all build install clean distclean format test
--- a/src/js/performance.js
+++ b/src/js/performance.js
@@ -13,7 +13,7 @@ class Performance {
     }
 
     now() {
-        return tjs.hrtimeBigInt() - this._startTime;
+        return Number(tjs.hrtimeBigInt() - this._startTime);
     }
 
     mark(name) {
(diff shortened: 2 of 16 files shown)


saghul:
Can we find a better solution for hrtime? I’d really like to have this, without having to call APIs like xxxBigInt

stevefan1999-personal:
@saghul the problem lies in here:
```js
// ... browserify shim
module.exports = process.hrtime || hrtime

// polyfil for window.performance.now
var performance = global.performance || {}
var performanceNow =
  performance.now        ||
  performance.mozNow     ||
  performance.msNow      ||
  performance.oNow       ||
  performance.webkitNow  ||
  function(){ return (new Date()).getTime() }

// generate timestamp or delta
// see http://nodejs.org/api/process.html#process_process_hrtime
function hrtime(previousTimestamp){
  var clocktime = performanceNow.call(performance)*1e-3 // ERROR: bigint cannot mix with number
  var seconds = Math.floor(clocktime)
  var nanoseconds = Math.floor((clocktime%1)*1e9)
  if (previousTimestamp) {
    seconds = seconds - previousTimestamp[0]
    nanoseconds = nanoseconds - previousTimestamp[1]
    if (nanoseconds<0) {
      seconds--
      nanoseconds += 1e9
    }
  }
  return [seconds,nanoseconds]
}
```
as you see performance.now() here implemented in performance.js returns a bigint, causing the program to crash on the way. And I think using a number can be justified because performance.now() returns a [DOMHighResTimeStamp](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp).

> The DOMHighResTimeStamp type is a **double** and is used to store a time value in milliseconds. This type can be used to describe a discrete point in time or a time interval (the difference in time between two discrete points in time).

Technically every number in JS is a de facto double precision floating number, I'm not sure why bigint is used instead.

saghul:
@stevefan1999-personal I see, thanks for explaining! I'll take a look.

saghul:
Hey there! Now that https://github.com/saghul/txiki.js/pull/105 landed, can you please undo the hrtime changes and rebase?

saghul:
Cherry-picked your commit and landed it in https://github.com/saghul/txiki.js/pull/113 cheers!

PR #98, opened 2020-01-12
Refactor argument parsing

vladimyr:
My C skill is really rusty but I did my best to abstract away all argparsing intricacies and to support GNU style long options with an equal sign: `--eval="console.log('Hello world.')"`


.clang-format
```diff
@@ -1,49 +1,49 @@
-﻿---
```
vladimyr: Removed `BOM` here. :bowtie:

.clang-format
```diff
@@ -1,49 +1,49 @@
...
+---
 AlignAfterOpenBracket: Align
-AlignConsecutiveMacros: 'true'
-AlignConsecutiveAssignments: 'false'
-AlignConsecutiveDeclarations: 'false'
+AlignConsecutiveMacros: true
```
vladimyr: According to schema clang format expects booleans/numbers instead of strings: http://clang.llvm.org/docs/ClangFormatStyleOptions.html#configurable-format-style-options

saghul:
This looks really nice, thanks! I intend to make some updates to the project this week, so I'll take a closer look at this too!

vladimyr:
Sure, take your time 👍
Thanks, I tweaked it too many times but I believe that I'm finally satisfied with the end result 🙃

src/cli.c
```diff
@@ -28,153 +28,228 @@
...
-static int eval_buf(JSContext *ctx, const void *buf, int buf_len, const char *filename, int eval_flags) {
+#define OPT_PREFIX '-'
+#define OPT_ASSIGN '='
+
+#define is_longopt(opt, str) (opt.name && !strncmp(opt.name, str, opt.length))
+
+typedef struct CLIOption {
+    char key;
+    char *name;
+    unsigned length;
```
saghul: can you please make this a size_t ?

src/cli.c
```diff
@@ -28,153 +28,228 @@
...
+} CLIOption;
+
+typedef struct Flags {
+    unsigned interactive, empty_run, strict_module_detection;
```
saghul: Please put each field in a separate line. Also, we can use bool here.

src/cli.c
```diff
@@ -28,153 +28,228 @@
...
-    exit(1);
+static void print_help(void) {
+    printf("Usage: tjs [options] [file]\n"
+           "\n"
+           "Options:\n"
+           "  -v, --version                     print tjs version\n"
+           "  -h, --help                        list options\n"
+           "  -e, --eval EXPR                   evaluate EXPR\n"
+           "  -i, --interactive                 go to interactive mode\n"
+           "      --strict-module-detection     only run code as a module if its extension is \".mjs\"\n"
```
saghul: sin you're reformatting this, maybe align the long options with the short ones

src/cli.c
```diff
@@ -28,153 +28,228 @@
...
+    printf("v%s\n", tjs_version());
+}
+
+static void report_bad_option(char *name) {
+    fprintf(stderr, "tjs: bad option -%s\n", name);
+}
+
+static void report_missing_argument(CLIOption *opt) {
+    if (opt->key)
+        fprintf(stderr, "tjs: -%c requires an argument\n", opt->key);
```
saghul: Maybe we should have a define like PROG_NAME instead of repeting tjs all the time...

src/cli.c
```diff
@@ -28,153 +28,228 @@
...
+}
+
+static unsigned get_option_length(const char *arg) {
+    const char *val_start = strchr(arg, OPT_ASSIGN);
+    if (!val_start)
+        val_start = arg + strlen(arg);
+    return val_start - arg - 1;
+}
+
+static int get_option(char **arg, CLIOption *opt) {
```
saghul: bool

src/cli.c
```diff
@@ -28,153 +28,228 @@
...
             break;
-        optind++;
-        if (*arg == '-') {
-            longopt = arg + 1;
-            arg += strlen(arg);
-            /* -- stops argument scanning */
-            if (!*longopt)
-                break;
+        optind += 1;
+        /* combining short options is NOT supported */
```
saghul: why not? it could make sense to use the prefload option in conjuction with the eval expr option.

src/cli.c
```diff
@@ -28,153 +28,228 @@
...
-                fprintf(stderr, "tjs: missing expression for --override-filename\n");
-                exit(2);
+                report_missing_argument(&opt);
+                exit_code = EXIT_INVALID_ARG;
+                goto exit;
             }
-            if (opt == 'i' || !strcmp(longopt, "interactive")) {
-                interactive++;
-                continue;
+            if (is_longopt(opt, "override-filename") || is_longopt(opt, "overrideFilename")) {
```
saghul: why the extra spelling? let's just have `override-filename`

src/cli.c
```diff
@@ -28,153 +28,228 @@
...
-                continue;
+            if (opt.key == 'q' || is_longopt(opt, "quit")) {
+                flags.empty_run = true;
+                break;
             }
-            if (opt) {
-                fprintf(stderr, "tjs: unknown option '-%c'\n", opt);
-            } else {
-                fprintf(stderr, "tjs: unknown option '--%s'\n", longopt);
+            if (is_longopt(opt, "strict-module-detection") || is_longopt(opt, "strictModuleDetection")) {
```
saghul: ditto

src/cli.c
```diff
@@ -28,153 +28,228 @@
...
         }
-        TJS_Run(qrt);
     }
 
-    TJS_FreeRuntime(qrt);
-    return 0;
-fail:
-    TJS_FreeRuntime(qrt);
-    return 1;
+    if (flags.interactive)
```
saghul: Add braces (I know they are not needed, but conistency!). I tried to force thi with clang-format, but I don't know how :-/

src/cli.c
```diff
@@ -28,153 +28,228 @@
...
+        TJS_RunRepl(ctx);
+    TJS_Run(qrt);
+
+exit:
+    if (qrt)
```
saghul: ditto

src/cli.c
vladimyr: I wanted to have one per line but clang format keeps bringing them together because line does not exceed certain limit.

src/cli.c
vladimyr: Combining in sense of joining them together: `-h -v` becomes `-hv`. (Given example does not have sense but you'll get the idea.) Why? Because it is damn hard to implement it and it even node's cli doesn't support it so I didn't waste time trying...

src/cli.c
saghul: Add a disable directive if need be

src/cli.c
saghul: Ah, gotcha. Ignore my comment then.

src/cli.c
vladimyr: Just for sake of convenience. That weird pascal casing arg style got popular after yarn introduced it. My thinking was it is easy enough to implement and makes everybody feel at home regardless of their background/expectations.

src/cli.c
vladimyr: Ok, let's learn how to force this using clang.

vladimyr:
> Thanks for this! It makes the code easier to read indeed! Left some comments, please take a look.

Thanks for your review! Making it more readable was my primary intention so hearing it back makes me really happy. 🤗

src/cli.c
saghul: I'd rather not have it :-)

vladimyr: Ok, I'll remove it.

src/cli.c
vladimyr: Or not, seems there is no way 🤦‍♂

saghul:
I'm ready to merge this. Are you thinking about making more changes?

vladimyr:
> I'm ready to merge this. Are you thinking about making more changes?

Nope, good to go 🚀

PR #99, opened 2020-01-12
Preload modules

vladimyr:
This is stacked onto #98 and enables module preloading like node's `-r, --require` flag does.

### DIFF
--- a/src/cli.c
+++ b/src/cli.c
@@ -70,9 +70,10 @@ static int eprintf(const char *format, ...) {
     return ret;
 }
 
-static int get_eval_flags(const char *filepath, unsigned strict_module_detection) {
-    if (strict_module_detection && !has_suffix(filepath, ".mjs"))
+static int get_eval_flags(const char *filepath, bool strict_module_detection) {
+    if (strict_module_detection && !has_suffix(filepath, ".mjs")) {
         return JS_EVAL_TYPE_GLOBAL;
+    }
     return JS_EVAL_TYPE_MODULE;
 }
 
@@ -229,9 +230,9 @@ int main(int argc, char **argv) {
                     exit_code = EXIT_INVALID_ARG;
                     goto exit;
                 }
-                FileItem *file = (FileItem *) (malloc(sizeof(FileItem)));
+                FileItem *file = malloc(sizeof(*file));
                 if (!file) {
-                    fprintf(stderr, "tjs: could not allocate memory\n");
+                    eprintf("could not allocate memory\n");
                     exit_code = EXIT_FAILURE;
                     goto exit;
                 }
@@ -292,7 +293,7 @@ int main(int argc, char **argv) {
         /* interactive mode */
         flags.interactive = true;
     } else {
-        char *filepath = argv[optind];
+        const char *filepath = argv[optind];
         int eval_flags = get_eval_flags(filepath, flags.strict_module_detection);
         if (eval_module(ctx, filepath, flags.override_filename, eval_flags)) {
             exit_code = EXIT_FAILURE;
--- a/src/cli.c
+++ b/src/cli.c
@@ -47,10 +47,16 @@ typedef struct CLIOption {
     size_t length;
 } CLIOption;
 
+typedef struct FileItem {
+    struct list_head link;
+    char *path;
+} FileItem;
+
 typedef struct Flags {
     bool interactive;
     bool empty_run;
     bool strict_module_detection;
+    struct list_head preload_modules;
     char *eval_expr;
     char *override_filename;
 } Flags;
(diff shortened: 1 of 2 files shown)


src/cli.c
```diff
@@ -28,153 +28,265 @@
...
+                char *filepath = get_option_value(arg, argc, argv, &optind);
+                if (!filepath) {
+                    report_missing_argument(opt, longopt);
+                    exit_code = EXIT_INVALID_ARG;
+                    goto exit;
                 }
-                if (optind < argc) {
-                    override_filename = argv[optind++];
-                    break;
+                FileItem *file = (FileItem *) (malloc(sizeof(FileItem)));
```
vladimyr: Forgot to free those later on. I got spoiled by garbage collector 🤦

saghul:
#98 landed (🚀 ) can you please rebase?

vladimyr:
> #98 landed (🚀 ) can you please rebase?

✅ Done.

src/cli.c
```diff
@@ -208,6 +222,23 @@ int main(int argc, char **argv) {
...
+            if (opt.key == 'l' || is_longopt(opt, "load")) {
+                char *filepath = get_option_value(arg, argc, argv, &optind);
+                if (!filepath) {
+                    report_missing_argument(&opt);
+                    exit_code = EXIT_INVALID_ARG;
+                    goto exit;
+                }
+                FileItem *file = (FileItem *) (malloc(sizeof(FileItem)));
+                if (!file) {
+                    fprintf(stderr, "tjs: could not allocate memory\n");
```
vladimyr: This should be altered to use `eprintf` instead.

src/cli.c
```diff
@@ -64,6 +70,12 @@ static int eprintf(const char *format, ...) {
     return ret;
 }
 
+static int get_eval_flags(const char *filepath, unsigned strict_module_detection) {
+    if (strict_module_detection && !has_suffix(filepath, ".mjs"))
```
saghul: style: please add braces

src/cli.c
```diff
@@ -97,6 +109,7 @@ static void print_help(void) {
            "  -v, --version                 print tjs version\n"
            "  -h, --help                    list options\n"
            "  -e, --eval EXPR               evaluate EXPR\n"
+           "  -l, --load FILENAME           module to preload (option can be repeated)\n"
```
saghul: can you check what other runtimes have as the flag? QuickJS just add -I --include

src/cli.c
```diff
@@ -208,6 +222,23 @@ int main(int argc, char **argv) {
...
                 goto exit;
             }
+            if (opt.key == 'l' || is_longopt(opt, "load")) {
+                char *filepath = get_option_value(arg, argc, argv, &optind);
+                if (!filepath) {
+                    report_missing_argument(&opt);
+                    exit_code = EXIT_INVALID_ARG;
+                    goto exit;
+                }
+                FileItem *file = (FileItem *) (malloc(sizeof(FileItem)));
```
saghul: No need to cast the result of malloc. Also, you can simplify this as follows:
`FileItem *file = malloc(sizeof(*file));`

src/cli.c
saghul: Also, I wonder if we should support passing a single option, with comma separated items. WDYT?

vladimyr: I don't think that is a good idea because filenames are allowed to contain both comma and whitespace characters making it impossible to disambiguate between _exotic_ filenames and list entry separators. Apart from that, it is pretty non-standard and I don't see the particular benefit over just going with repeated flags.

vladimyr: `node` uses `-r, --require`: https://nodejs.org/api/cli.html#cli_r_require_module because it basically does `require()` call under the hood while in our case we are dealing with esm imports. OTOH I couldn't find any equivalent `deno` flag. Haven't used other runtimes so my insights are quite limited 🤷‍♂

saghul: 👍

src/cli.c
```diff
@@ -64,6 +70,13 @@ static int eprintf(const char *format, ...) {
     return ret;
 }
 
+static int get_eval_flags(const char *filepath, unsigned strict_module_detection) {
```
saghul: bool strict_module_detection

src/cli.c
saghul: qjs and tjs are nothing alike at this point, so let's go with your proposal!

src/cli.c
```diff
@@ -249,10 +293,8 @@ int main(int argc, char **argv) {
         /* interactive mode */
         flags.interactive = true;
     } else {
-        const char *filepath = argv[optind];
-        int eval_flags = JS_EVAL_TYPE_MODULE;
-        if (flags.strict_module_detection && !has_suffix(filepath, ".mjs"))
-            eval_flags = JS_EVAL_TYPE_GLOBAL;
+        char *filepath = argv[optind];
```
saghul: i think you can keep the const here

PR #110, opened 2020-01-28
core: fix promise microtask running sequence

### LINKED ISSUE #107: Issue on Async Running Sequence
It seems that txiki's event loop doesn't follow the microtask / macrotask concepts strictly. For example:

``` js
console.log('Hello')

setTimeout(() => console.log('!'), 0)

Promise.resolve().then(() => console.log('World'))
```

This snippet has following output in Chrome's console and `qjs`:

```
Hello
World
!
```

Since `Promise.then` returns a microtask, and if my understanding is correct, it should be executed by `JS_ExecutePendingJob` synchronously right after each eval. So the log from `Promise.then` comes before the one from `setTimeout`.

While in txiki I got following output:

```
Hello
!
World
```

Currently microtasks are triggered by `uv_check` after each tick, while after the first time we eval the script, there isn't corresponding `JS_ExecutePendingJob` calls, so the `Promise`'s microtask is kept until next tick's check phase in `setTimeout`'s macrotask, which makes the difference in running sequence.

doodlewind:
Fix #107 

We don't have to migrate to `UV_RUN_ONCE` for this case. This PR simply executes the microtask jobs before the `setTimeout` callback, which makes the `Promise.then` microtasks can be always cleared before `setTimeout`.

I've verified the basic and the nesting callback case in the original issue.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -259,15 +259,15 @@ static void uv__prepare_cb(uv_prepare_t *handle) {
 }
 
 void tjs_execute_jobs(JSContext *ctx) {
-    JSRuntime *rt = JS_GetRuntime(ctx);
+    JSContext *ctx1;
     int err;
 
     /* execute the pending jobs */
     for (;;) {
-        err = JS_ExecutePendingJob(rt, &ctx);
+        err = JS_ExecutePendingJob(JS_GetRuntime(ctx), &ctx1);
         if (err <= 0) {
             if (err < 0)
-                tjs_dump_error(ctx);
+                tjs_dump_error(ctx1);
             break;
         }
     }
--- a/src/vm.c
+++ b/src/vm.c
@@ -258,23 +258,26 @@ static void uv__prepare_cb(uv_prepare_t *handle) {
     uv__maybe_idle(qrt);
 }
 
-static void uv__check_cb(uv_check_t *handle) {
-    TJSRuntime *qrt = handle->data;
-    CHECK_NOT_NULL(qrt);
-
-    JSRuntime *rt = qrt->rt;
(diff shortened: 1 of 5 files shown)


saghul:
Hum, interesting way of solving it! Can you please DRY the execute_jobs function and expose it in private.h so we can use it in both places we need it?

saghul:
Also, if you could add a test that would be ideal.

src/vm.c
```diff
@@ -258,23 +258,26 @@ static void uv__prepare_cb(uv_prepare_t *handle) {
...
-    JSRuntime *rt = qrt->rt;
-    JSContext *ctx1;
+void tjs_execute_jobs(JSContext *ctx) {
+    JSRuntime *rt = JS_GetRuntime(ctx);
     int err;
 
     /* execute the pending jobs */
     for (;;) {
-        err = JS_ExecutePendingJob(rt, &ctx1);
+        err = JS_ExecutePendingJob(rt, &ctx);
```
saghul: THis is not correct. Please leave ctx1 and take a TJSRuntime as an argument to tjs_execute_jobs. The reason is that JS_ExecutePendingJob returns the context that is resumed in the passed ctx argument. It need not be the one we have (if we ever go multi-context).

doodlewind: I can't get the `TJSRuntime` directly in the timer's callback. Current solution aligns with the `js_std_loop` pattern in QuickJS.

doodlewind:
@saghul timer tests are also ready 😀

saghul:
@doodlewind Awesome work, thank you! 🚀

PR #120, opened 2020-02-19
Fix segfault by twice calling curl_easy_cleanup() at async mode.

junsik:


junsik:
Sorry, I found a bug. :(

saghul:
No worries. Can you please remove the first commit? It already landed and it causes a conflict. Bonus points if you add a regression test!

junsik:
Sorry for late response. I am not used to github.
It looks like fixed. But travis-ci build failed by test-performance.js. :(
Do I need to close this and re-create the pull-request again?

saghul:
No worries, that test is a bit flaky. Now, any chance you can add a test which crashes without your fix but passes with it?

saghul:
I landed your patch in ca1890b, cheers!

Issue #117, opened 2020-02-02
Setting max stack size limit


cabrrtt:
Would be great to be able to set max stack size limit via the command line / env var

saghul:
What use case do you have in mind? Are you running into problems with latest master? (I just raised it  in the last few days)

cabrrtt:
Not running into any issues no. I spotted that you had raised it in a recent commit. It occurred to me it would be helpful to be able to set this a startup so as to avoid excessive memory growth, with stack overflow error indicating offending call.

saghul:
I think this API is what you're really after, isn't it? https://github.com/saghul/txiki.js/blob/d1b5d9f636856fdcda2d5293e8ad3dba786e4e04/deps/quickjs/include/quickjs.h#L334

cabrrtt:
Indeed it is. Thanks

saghul:
I wouldn't mind adding that. Are you up for writing a PR?

rickyes:
#122  Already supports `--stack-size` command.

saghul:
I this issue is really more about supporting a new memory limit option.

PR #122, opened 2020-03-02
feat: add SetMaxStackSize

rickyes:


saghul:
Did you make changes so this is necessary? Note that this is the auxiliary tool to bundle JS files, the tjs binary already has an increased stack size.

rickyes:
Hi，My idea is to change the size of the runtime stack with the command line argument

saghul:
Yes, but you are changing it in qjsc.c, which is not the main interpreter. It's an auxiliary tool to create the bytecode that is then embedded in tjs. You may want to start looking at cli.c and vm.c.

rickyes:
It has been modified.

src/cli.c
```diff
@@ -248,6 +250,11 @@ int main(int argc, char **argv) {
                 exit_code = EXIT_INVALID_ARG;
                 goto exit;
             }
+            if (is_longopt(opt, "stack-size")) {
+                char *stack_size = get_option_value(arg, argc, argv, &optind);
+                runOption.stack_szie = (size_t) stack_size;
```
saghul: you probably need `atoi` here, you can't just cast the string to a size_t

src/private.h
```diff
@@ -63,6 +63,10 @@ struct TJSRuntime {
 #endif
 };
 
+typedef struct RunOption {
```
saghul: This shouldn't be private, let's move it to tjs.h

src/private.h
```diff
@@ -63,6 +63,10 @@ struct TJSRuntime {
...
+    size_t stack_szie;
+} RunOption;
```
saghul: let's call this TJSRunOptions

src/tjs.h
```diff
@@ -34,8 +34,9 @@
 
 typedef struct TJSRuntime TJSRuntime;
 
-TJSRuntime *TJS_NewRuntime(void);
+TJSRuntime *TJS_NewRuntime(RunOption *option);
 TJSRuntime *TJS_NewRuntime2(bool is_worker);
+TJSRuntime *TJS_NewRuntimeByOptions(bool is_worker, RunOption *option);
```
saghul: Please leave `TJSRuntime *TJS_NewRuntime(void);` alone, taking the default values.

saghul: Let's call this TJS_NewRuntime2 instead and skip the `is_worker` parameter. I'll make the current TJS_NewRuntime2 private. You can rename it to TJS_NewRuntimeWorker in the interim.

src/vm.c
```diff
@@ -27,6 +27,8 @@
 
 #include <string.h>
 
+#define MAX_STACK_SIZE 1048576
```
saghul: `TJS__DEFAULT_STACK_SIZE`

src/tjs.h
rickyes: I plan to change it like this, what do you think:

``` c
TJSRuntime *TJS_NewRuntime(void);
TJSRuntime *TJS_NewRuntimeWorker(void);
TJSRuntime *TJS_NewRuntimeOption(bool is_worker, TJSRunOptions *option);
```

saghul: Sounds good!

saghul:
Nice work, thank you very much!

rickyes:
#117

PR #140, opened 2020-04-09
utils: do JS_FreeValue at the right place

topelinux:
Maybe `JS_FreeValue` should be called at `TJS_InitPromise` for `rfuncs`

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -185,6 +185,8 @@ JSValue TJS_InitPromise(JSContext *ctx, TJSPromise *p) {
         return JS_EXCEPTION;
     p->rfuncs[0] = JS_DupValue(ctx, rfuncs[0]);
     p->rfuncs[1] = JS_DupValue(ctx, rfuncs[1]);
+    JS_FreeValue(ctx, rfuncs[0]);
+    JS_FreeValue(ctx, rfuncs[1]);
     return JS_DupValue(ctx, p->p);
 }
 
@@ -221,8 +223,6 @@ void TJS_SettlePromise(JSContext *ctx, TJSPromise *p, bool is_reject, int argc, 
     for (int i = 0; i < argc; i++)
         JS_FreeValue(ctx, argv[i]);
     JS_FreeValue(ctx, ret); /* XXX: what to do if exception ? */
-    JS_FreeValue(ctx, p->rfuncs[0]);
-    JS_FreeValue(ctx, p->rfuncs[1]);
     TJS_FreePromise(ctx, p);
 }
 


saghul:
I don't think so. DupValue / FreeValue are for incrementing / decrementing the JSValue's refcount. Why would you be decrementing the refcount you just incremented?

topelinux:
Thanks for you response!

Let me update my founding
The patch i though should be

```c
diff --git a/src/utils.c b/src/utils.c
index cee0252..2d50e88 100644
--- a/src/utils.c
+++ b/src/utils.c
@@ -183,8 +183,9 @@ JSValue TJS_InitPromise(JSContext *ctx, TJSPromise *p) {
     p->p = JS_NewPromiseCapability(ctx, rfuncs);
     if (JS_IsException(p->p))
         return JS_EXCEPTION;
-    p->rfuncs[0] = JS_DupValue(ctx, rfuncs[0]);
-    p->rfuncs[1] = JS_DupValue(ctx, rfuncs[1]);
+    p->rfuncs[0] = rfuncs[0];
+    p->rfuncs[1] = rfuncs[1];
+
     return JS_DupValue(ctx, p->p);
 }

@@ -221,8 +222,6 @@ void TJS_SettlePromise(JSContext *ctx, TJSPromise *p, bool is_reject, int argc,
     for (int i = 0; i < argc; i++)
         JS_FreeValue(ctx, argv[i]);
     JS_FreeValue(ctx, ret); /* XXX: what to do if exception ? */
-    JS_FreeValue(ctx, p->rfuncs[0]);
-    JS_FreeValue(ctx, p->rfuncs[1]);
     TJS_FreePromise(ctx, p);
 }

```
No need do `DupValue` for ` p->rfuncs[0].rfuncs` at `TJS_InitPromise` , for you already obtain the ownership by `rfuncs[0]` you just transfer the ownership to `p->rfuncs[0].rfuncs`.

Then at the `TJS_SettlePromise` stage, don't need to `JS_FreeValue(ctx, p->rfuncs[0])`,  for it just release the ownership at  ` TJS_FreePromise(ctx, p)`

That is my understanding, thanks for comments.

But during my verification, when trigger unit test run.js in `tests`, it could pass for `Release` build, but panic for `Debug` build.
So i doubt there is something wrong of my understanding, or that are other issue around elsewhere.

saghul:
> But during my verification, when trigger unit test run.js in `tests`, it could pass for `Release` build, but panic for `Debug` build.
> So i doubt there is something wrong of my understanding, or that are other issue around elsewhere.

I'll admit I struggled there myself too. If you look at the commit history around those lines you'll see it took me some iterations to make sure tests passed and things didn't crash. There might a bug hidden somewhere, but honestly I haven't been able to find it.

PR #148, opened 2020-10-10
fix: excuse idle_cb when no pending job

DingSoung:
JS_ExecutePendingJob return "0" means  list_empty true
while JS_IsJobPending return "true" means list_empty false


saghul:
This is incorrect. uv_idle_t has a very bad name, it doesn't run when the loop is idle, on the contrary, it prevents the uv event loop for blocking for i/o. Thus, the usage here is correct: when there are pending jobs, we enable an empty uv_idle_t handle which will prevent the loop from blocking for i/o.

PR #152, opened 2020-11-10
deps: update curl, libuv and replxx to their latest versions

saghul:


tritao:
With QuickJS having an official repo on Github now (https://github.com/bellard/quickjs) could it be changed to point to a submodule as well or would there be any issues with it?

PR #179, opened 2021-10-09
optionally build a shared library

brandonros:
I'm not very good at CMake so let me know how this can be better.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,6 +100,31 @@ add_executable(tjs
     ../deps/quickjs/src/cutils.c
 )
 
+add_library(tjs SHARED
+    src/bootstrap.c
+    src/cli.c
+    src/curl-utils.c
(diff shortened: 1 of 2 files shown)


brandonros:
I'm not quite sure what the "on library load" function should be, let alone the `named exports` we need to now export. `main()` would need to be converted, and we'd need a way to access "should-be-globals" like `TJSRuntime *qrt and JSContext *ctx`

brandonros:
```
$ nm build/libtjs.dylib | grep TJS_GetJSContext
000000000000fb84 T _TJS_GetJSContext
$ nm build/libtjs.dylib | grep TJS_GetRuntime
000000000000fb8c T _TJS_GetRuntime
```

Not bad.

KaruroChori:
In theory https://github.com/saghul/txiki.js/pull/517 made this redundant right?

PR #181, opened 2021-11-12
Update to latest quickjs release

johnrobinsn:
* Update to latest quickjs release 2021-03-27
* generate a static library 'quv' for the core libuv runtime support for embedding purposes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,9 +75,8 @@ if(NOT DISABLE_CURL)
     endif()
 endif()
 
-add_executable(tjs
+add_library(quv STATIC
     src/bootstrap.c
-    src/cli.c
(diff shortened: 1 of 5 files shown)


saghul:
Thanks! Can you please break this into 2 PRs? We do want to update, but I'm not yet sure about the extra target.

johnrobinsn:
Ok.  I'm embedding the quv lib into another project.  So I think a useful addition.  But will split them apart.

saghul:
Right. In that case let's have it, but separate please. It should be called libtjs, since quv was a short lived previous name this project had.

johnrobinsn:
Makes sense... closing this PR and will open new ones

saghul:
Thank you!

PR #184, opened 2021-11-14
Upstream qjs debugger clean

johnrobinsn:
Integrated quickjs debugger support from https://github.com/koush/vscode-quickjs-debug

Adds the ability to debug javascript with Visual Studio Code (using a quickjs debugger extension).

QuickJS Debugger supports stepping, continue, breakpoints, evaluation, and variable access.

QuickJS Debugger support is not compiled in by default. Given tight integration within the quickjs opcode dispatcher and unsure of performance implications.

To enable support
mkdir build; cd build; cmake .. -DCONFIG_QUICKJS_DEBUGGER=ON; make


saghul:
Did you intend to close this?

johnrobinsn:
Sorry.... Didn't seem like there was interest... Reopened.

saghul:
There is interest, sorry! What I don't have much of is time :-)

saghul:
It's still a huge change, but I don't see any movement in upstream QuickJS so yeah.

It does need a rebase though.

deps/quickjs/src/quickjs.c
```diff
@@ -16226,10 +16264,30 @@ static JSValue JS_CallInternal(JSContext *caller_ctx, JSValueConst func_obj,
...
+#include "quickjs-opcode.h"
+        [ OP_COUNT ... 255 ] = &&case_default
+    };
+#define SWITCH(pc)      goto *active_dispatch_table[opcode = *pc++];
+#define CASE(op)        case_debugger_ ## op: js_debugger_check(ctx, pc); case_ ## op
+#define DEFAULT         case_default
+#define BREAK           SWITCH(pc)
+
+    const void * const * active_dispatch_table = caller_ctx->rt->debugger_info.transport_close
+        ? debugger_dispatch_table : dispatch_table;
```
bnoordhuis, line 16289: Doesn't adding this check plus several pointer indirections introduce pretty hefty overhead to bytecode dispatch? quickjs bytecodes are "fat" but dispatch is often still 10-20% of running time and this change likely makes it worse. Maybe run some before/after benchmarks?

On a side note: I believe SHORT_OPCODES=0 and OPTIMIZE=0 are broken and DIRECT_DISPATCH=0 (emscripten support) probably isn't relevant to txiki so maybe the ifdef-ery here can be simplified somewhat.

Issue #247, opened 2022-03-02
More FS operations


saghul:
fs operations:

Done!

samuelbrian:
Sorry about all the churn here. I wasn't expecting pushing to my forked repo to show here.

saghul:
No worries at all mate, thanks for your contributions!

PR #281, opened 2022-04-19
fs: add mkdir()

samuelbrian:
Hi,

I've come across the need for mkdir so I implemented it and it is presented here in this PR.

But I was hoping to improve on the usability of the `mode` parameter, considering octal literals are deprecated, but I'm not sure what would fit best.

In the TypeScript definition I see the `S_XXX` variable, but I can't find any of the flags mentioned added anywhere. If I've just missed them somehow, please let me know.

First thought was to declare them in fs.c e.g.
```
--- a/src/fs.c
+++ b/src/fs.c
@@ -1151,6 +1151,29 @@ static const JSCFunctionListEntry tjs_fs_funcs[] = {
     TJS_CONST2("COPYFILE_EXCL", UV_FS_COPYFILE_EXCL),
     TJS_CONST2("COPYFILE_FICLONE", UV_FS_COPYFILE_FICLONE),
     TJS_CONST2("COPYFILE_FICLONE_FORCE", UV_FS_COPYFILE_FICLONE_FORCE),
+    TJS_CONST(S_IFMT),
+    TJS_CONST(S_IFSOCK),
+    TJS_CONST(S_IFLNK),
+    TJS_CONST(S_IFREG),
...
+    TJS_CONST(S_IRWXO),
+    TJS_CONST(S_IROTH),
+    TJS_CONST(S_IWOTH),
+    TJS_CONST(S_IXOTH),
     TJS_CFUNC_DEF("open", 3, tjs_fs_open),
     TJS_CFUNC_DEF("newStdioFile", 2, tjs_fs_new_stdio_file),
     TJS_CFUNC_MAGIC_DEF("stat", 1, tjs_fs_stat, 0),

```
But that pollutes the tjs namespace a lot. There are over 20 of them.

To group them together I considered adding them as an enum to the .d.ts but that doesn't help people who don't use TypeScript.
Or maybe they could be added to an object under tjs in fs.c like `tjs.FileMode`, and declared as constants in .d.ts.

Any thoughts on what would be preferred here, if any?

Thanks,

Sam

-------------

Added tjs.mkdir() and simple test.
Mode mask parameter defaults to 0777, which is the default from the
coreutils mkdir program.

Ref: https://github.com/saghul/txiki.js/issues/247

### DIFF
--- a/src/fs.c
+++ b/src/fs.c
@@ -316,6 +316,7 @@ static void uv__fs_req_cb(uv_fs_t *req) {
         case UV_FS_FDATASYNC:
         case UV_FS_FSYNC:
         case UV_FS_FTRUNCATE:
+        case UV_FS_MKDIR:
         case UV_FS_RENAME:
         case UV_FS_RMDIR:
         case UV_FS_UNLINK:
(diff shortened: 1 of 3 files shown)


saghul:
Thanks for the PR! Looks good at a quick glance!

> considering octal literals are deprecated

Are they? I thoughts the 0-numbers where, but not is using the octal literal syntax `0o777` for example. Are those deprecated too?

samuelbrian:
> Are they?

I was referring to the 0 numbers and didn't even realise the `0o` notation existed! OK, that makes it a lot easier to use.

I suspect those check failures are related to the note here http://docs.libuv.org/en/v1.x/fs.html#c.uv_fs_mkdir about the mode not being supported.
I'll get on a Windows computer soon and check it out.

bnoordhuis:
> I suspect those check failures are related to the note here http://docs.libuv.org/en/v1.x/fs.html#c.uv_fs_mkdir about the mode not being supported on Windows.

Just to confirm: that's indeed what happens. Unix-style modes aren't a thing on Windows, therefore `uv_fs_mkdir()` ignores the `mode` argument, and `uv_fs_stat()` always sets `statbuf->st_mode` to 0.

samuelbrian:
Thanks @bnoordhuis.
I've changed the test to only check the permission bits if `tjs.platform` isn't `"windows"`.
To be honest, I haven't run it on Windows and hoping the automated checks will do it for me :smile:

src/fs.c
```diff
@@ -1131,6 +1161,7 @@ static const JSCFunctionListEntry tjs_fs_funcs[] = {
     TJS_CFUNC_DEF("mkdtemp", 1, tjs_fs_mkdtemp),
     TJS_CFUNC_DEF("mkstemp", 1, tjs_fs_mkstemp),
     TJS_CFUNC_DEF("rmdir", 1, tjs_fs_rmdir),
+    TJS_CFUNC_DEF("mkdir", 1, tjs_fs_mkdir),
```
saghul, line 1164: Say 2 here since it has 2 parameters.

saghul (commented):
One nit LGTM otherwise!

src/fs.c
samuelbrian: Done!

saghul (approved):

saghul:
Cheers!

PR #284, opened 2022-04-25
Added chmod, chown, lchown, kill

samuelbrian:
Hi,

I've added some more fs functions (chown, lchown, chmod) and the non-process-handle kill function.

The test for kill() on Windows is not very effective because of some limitations with that platform, but I dove into libuv and its own tests and it looks like it's to be expected.


saghul:
These look great, thank you!

saghul (approved):

PR #285, opened 2022-04-25
TypeScript definition file fixes and extras

samuelbrian:
I've been using TypeScript for a few scripts and have noticed a some parts txikijs.d.ts are out of sync with what tjs provides.

Instances of `tjs.Error` have the `errno` property and not the `code` property.
I've described `tjs.Error` as a class with static properties, but was this deliberately avoided?
I've listed the error numbers explicitly, because `EXXX` is not useful in TypeScript.
Considering the recent push to change signal names to strings, would it also be desirable to only expose the error name strings and not the error numbers?

For signals, I've added a `Signal` type to help with type checking when using the signal names.
I've also annotated the `tjs.signal()` return type and added a test for it because I couldn't work out how to use it until I realised that you need to prevent the returned handler from being garbage collected.

Is the definition file's main purpose to generate the docs currently? I think the best way to keep it in sync with the implementation would be to have the tests written in TypeScript.
If I converted the tests to TypeScript, would you be interested in the changes?

Let me know what you think.

### DIFF
--- a/docs/txikijs.d.ts
+++ b/docs/txikijs.d.ts
@@ -69,7 +69,7 @@ declare namespace tjs {
         /**
          * The signal that this signal handler was registered for.
          */
-        signum: number;
+        signal: Signal;
 
         /**
          * Stop the signal handler. The registered signal handler function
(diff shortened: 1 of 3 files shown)


saghul (approved):

saghul:
Great improvements thank you!

And yeah errnos should probably become strings.

PR #287, opened 2022-05-05
xhr: fix ProgressEvent parameters not being passed to onprogress

samuelbrian:
Hi @saghul,
This fixes XMLHttpRequest's onprogress not receiving the progress-related values in its event due to using the Event class directly.
It was the only occurrence I found of the Event constructor being used like this.


saghul (approved):
Nice one! I triggered a test re-run, though I think the failure is spurious.

PR #288, opened 2022-05-05
process: added exec(cmd) wrapper around execvp()

samuelbrian:
Hi @saghul,

I've been using tjs for scripting in an embedded Linux environment, and although there is no libuv helper function, being able to exec is pretty useful.

### DIFF
--- a/docs/txikijs.d.ts
+++ b/docs/txikijs.d.ts
@@ -792,6 +792,16 @@ declare namespace tjs {
      */
     function spawn(args: string | string[], options?: ProcessOptions): Process;
 
+    /**
+     * Replace the current process image with a new process image.
+     * This function does not return if successful.
+     *
+     * See [execvp(3)](https://man7.org/linux/man-pages/man3/execvp.3.html)
+     *
+     * @param args Command argument list for the new process image.
+     */
+    function exec(args: string | string[]): void;
+
     interface Address {
         family: number;
         ip: string;
--- a/src/process.c
+++ b/src/process.c
@@ -457,6 +457,68 @@ cleanup:
     return ret;
 }
 
+static JSValue tjs_exec(JSContext *ctx, JSValueConst this_val, int argc, JSValueConst *argv) {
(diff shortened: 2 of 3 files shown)


samuelbrian:
In short, I'm using tjs as an alternative to shell for system service scripts that read/write JSON.

I've got a program that runs system services, sort of like an init, and it keeps track of each service process's pid so that it can be sent signals.
These service processes are generally shell scripts, and more recently tjs scripts, which write configuration files and then start a daemon.

If the script spawns the daemon as a subprocess with the daemon foregrounded, then the script has to keep running and wait on the daemon and proxy signals to it.

If the script spawns the daemon as a subprocess with the daemon backgrounding itself, then it must exit and deal with pidfiles when it is run again to restart or SIGHUP the daemon.

If the script execs into the daemon (foregrounded), then I can take advantage of the killing logic in the system runner and avoid a process hanging around with such a minimal purpose.

I'm not sure if exec() is useful to all, but it's in enough of other language's runtimes/stdlib so I thought it might be a good fit.
No worries if you don't want to take it. I keep some patches for other things I expect have no chance of being accepted, like a more complicated module resolution that works sort of like nodejs and searches paths listed in the env variable `TJS_LIBRARY_PATH` which defaults to `/lib/js`.

saghul:
Would exposing the UV_PROCESS_DETACHED flag for spawn help? Or would that be case 2 in your list?

I'm a bit hesitant on this one since it opens the window for platform dependent stuff I'm not sure I want to add. I'll say I am surprised it just worked on MinGW :-)

Let me think about this one some more.

samuelbrian:
Yep, UV_PROCESS_DETACHED is case 2.

I sat on this change for a while because originally I was using execvpe() which wasn't available on MacOS or MinGW.
But since ditching the env setting and using execvp() and getting it working on all the CI platforms I thought it had a shot :)

saghul (approved):

saghul:
Let's have it.

Issue #291, opened 2022-05-08
read() and write() expect TypedArray as parameter in examples/process.js


guest271314:
The following code

https://github.com/saghul/txiki.js/blob/017cd0ad7efa9b5d59c4cb492ab0ea530f739408/examples/process.js#L37-L50

throws

```
 TypeError: not a TypedArray
    at write (native)
```
```
TypeError: not a TypedArray
    at read (native)
```

The code should be something like

```
#!tjs
(async() => {
    let data;
    let encoder = new TextEncoder();
    let decoder = new TextDecoder();
    let proc = tjs.spawn('cat', { stdin: 'pipe', stdout: 'pipe' });
    console.log(`proc PID: ${proc.pid}`);
    console.log(proc.stdin.fileno());
    console.log(proc.stdout.fileno());
    let input = encoder.encode('hello!');
    proc.stdin.write(input);
    data = new Uint8Array(input.length);
    await proc.stdout.read(data);
    console.log(decoder.decode(data));
    input = encoder.encode('hello again!');
    proc.stdin.write(input);
    data = new Uint8Array(input.length);
    await proc.stdout.read(data);
    console.log(decoder.decode(data));
    proc.kill(tjs.SIGTERM);
    let status = await proc.wait();
    console.log(status);
})()
.catch(console.error);
```

saghul:
Opsie. Care to send a PR?

guest271314:
I'll send a PR. Is the example substitute code sufficient as-is?

There are some other issues that need to be fixed, e.g., this URL https://github.com/saghul/txiki.js/blob/master/docs/modules/tjs.html is 404 at https://github.com/saghul/txiki.js/blob/master/docs/intro.md.

Additionally, I am having issues adapting this code https://github.com/simov/native-messaging/blob/master/protocol.js, https://github.com/simov/native-messaging/blob/master/nodejs/example.js to txiki.js (and QuickJS) as a Native Messaging host - when `tjs.spawn()` (`popen()` in QuickJS) is used; I am still testing to make sure it is not my code that is the problem.

PR #292, opened 2022-05-08
read() and write() expect TypedArray as parameter

guest271314:
Fixes https://github.com/saghul/txiki.js/issues/291


saghul:
Thanks!

PR #297, opened 2022-05-25
fs: default tjs.open() mode parameter to 0666

samuelbrian:
When opening files for creation, it is rarely desirable to have no file
mode bits set.
This change sets the default file mode bits to 0666, which is what NodeJS
and shells set it to, allowing the umask to reduce the permissions
appropriately.


saghul:
Cheers!

PR #299, opened 2022-07-05
Added ffi

lal12:
Hey this is my first go at adding FFI to txiki. It is mostly a copy from https://github.com/shajunxing/quickjs-ffi.

There are some improvements to be done for ts types and more tests.
But wanted to check whether the general approach works for you, e.g. adding ffi as a builtin module (`@tjs/ffi`).


saghul:
Any chance we could add libffi as a submodule instead of using a dependency? A dream of mine to eventually have all dependencies bundled in.

lal12:
Added libffi as a submodule, though this was a bit fiddly calling an autotools project from CMake

ffi is now available under tjs.ffi. Though I am not convinced that this is the best way, it could get cluttered fast...

lal12:
I hadn't really used the [shajunxing/quickjs-ffi](https://github.com/shajunxing/quickjs-ffi) api before. I am not that happy about that. Have seen better examples for ffi apis. I am currently considering either improving the API on the JS side or even rewriting some/most of the C code. So anyway there might still be larger things to worry about first.

Will get around to doing/deciding that hopefully next week.

saghul:
Given that we use Node to generate the bundle I'd say introducing TS should be fine.

The only concern could perhaps be source maps.

lal12:
So I've added my own implementation of a QuickJS wrapper. 

Basic functionality is there. 

Still missing:
- ts types
- struct types are not tested and have not JS code yet
- JS callbacks to C functions are not yet implemented

saghul:
Nice!

lal12:
- completed struct types
- added JSCallbacks
- added little C protoype parser for easier use of FFI api. It probably won't work for any kind of crazy c header (like e.g. libc), but most "normal" prototypes are supported.

So I guess/hope/think I'm done. Onlything missing are the TS Types, which I will add tomorrow.

saghul:
Awesome news! I'll take a look ASAP.

lal12:
Ok CI mostly works now, but:

MacOS sometimes (not always) fails in test-performance.js. Maybe the github actions MacOS vm is just a bit slow...

Windows just won't build libffi, with some weird error. So for windows I changed to the external libffi for now.

I added esbuild as a Command in Cmake, however apparently msys2 doesn't have nodejs, so it can't be run. I mostly did it for convenience in my dev process, so I could change that back. Though I think it is generally good not to have the build results (like std.js or bundle.js) committed in the repo, so I would propose bundling them in a pre step in a linux container?

.github/workflows/ci.yml
```diff
@@ -1,6 +1,6 @@
 name: CI
 
-on: [pull_request]
+on: [push, pull_request]
```
saghul, line 3: Why is this necessary?

CMakeLists.txt
```diff
@@ -75,12 +57,39 @@ add_subdirectory(deps/wasm3 EXCLUDE_FROM_ALL)
...
+set(INCLUDED_JS_FILES_PATH "")
+foreach(jsfile ${INCLUDED_JS_FILES})
+    configure_file(
+        ${CMAKE_SOURCE_DIR}/src/js/${jsfile}
+        ${CMAKE_CURRENT_BINARY_DIR}/${jsfile}
+        COPYONLY)
+	list(APPEND INCLUDED_JS_FILES_PATH ${CMAKE_CURRENT_BINARY_DIR}/${jsfile})
+endforeach()
+
+add_custom_command(
```
saghul, line 72: I appreciate the effort here, but I'd rather discuss changing the way we bundle things as a separate effort.

The main reason for how things are this way is to require virtually nothing other than a compiler on the target OS.

I am happy to re-evaluate, but not intermingled with FFI, which is already a huge change.

Can you please re-generate and commit the bundle on this PR, and open an issue so we discuss making this change afterwards? Thank you!

Dockerfile
```diff
@@ -1,11 +1,11 @@
 FROM alpine:latest
 ENV VERSION 20191124
 WORKDIR /
-RUN apk add build-base cmake curl-dev git --update-cache
+RUN apk add build-base cmake curl-dev ffi-dev git --update-cache
```
saghul, line 4: Weren't we going to use the embedded ffi on Linux and macOS but not on Windows?

src/ffi.c
```diff
@@ -0,0 +1,1103 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*/
+
+#include "private.h"
+
+#include <ffi.h>
+#include <stdint.h>
+
+#define C_MACRO_STRING_DEF(x) JS_PROP_STRING_DEF(#x, x, JS_PROP_CONFIGURABLE)
```
saghul, line 30: Generally I've been using `TJS__` for custom macros that we define as helpers. Just to make it obvious that the name doesn't represent something any file expects.

src/ffi.c
```diff
@@ -0,0 +1,1103 @@
...
+#define ffi_type_ptr ffi_type_uint32
+#elif UINTPTR_MAX == UINT64_MAX
+#define JS_TO_UINTPTR_T(ctx, pres, val) JS_ToBigInt64(ctx, (int64_t *)(pres), val)
+#define JS_NEW_UINTPTR_T(ctx, val) JS_NewBigUint64(ctx, (int64_t)(val))
+#define ffi_type_ptr ffi_type_uint64
+#else
+#error "'uintptr_t' neither 32bit nor 64 bit, I don't know how to handle it."
+#endif
+
+#define STR(x) #x
```
saghul, line 47: This is in utils.h IIRC.

src/ffi.c
```diff
@@ -0,0 +1,1103 @@
...
+#error "'int' neither 32bit nor 64 bit, I don't know how to handle it."
+#endif
+
+#define FFI_ALIGN(v, a)  (((((size_t) (v))-1) | ((a)-1))+1)
+
+#pragma region "FFI Helpers"
+// ===================
+
+static const char* ffi_strerror(ffi_status status){
+    switch(status){
```
saghul, line 85: are there any other error cases? should we have a `default` clause which returns "FFI_UNKNOWN" perhaps?

saghul, line 85: Can you run clang format? I think it will get rid of some of the style issues I am seeing, like missing spaces here.

src/ffi.c
```diff
@@ -0,0 +1,1103 @@
...
+
+#if SIZE_MAX == UINT32_MAX
+    ADD_ALIAS_TYPE(ctx, ffiobj, type_size, type_uint32);
+    ADD_ALIAS_TYPE(ctx, ffiobj, type_ssize, type_sint32);
+#else
+    ADD_ALIAS_TYPE(ctx, ffiobj, type_size, type_uint64);
+    ADD_ALIAS_TYPE(ctx, ffiobj, type_ssize, type_sint64);
+#endif
+
+    //C_VAR_ADDRESS_DEF(ffi_type_complex_float),
```
saghul, line 1100: Is this a TODO or can the comment go away?

src/ffi.c
```diff
@@ -0,0 +1,1103 @@
...
+#define C_ENUM_DEF(x) JS_PROP_INT64_DEF(#x, (int64_t)(x), JS_PROP_CONFIGURABLE)
+#define JS_TO_INT(ctx, pres, val) JS_ToInt64(ctx, (int64_t *)(pres), val)
+#define JS_NEW_INT(ctx, val) JS_NewInt64(ctx, (int64_t)(val))
+#else
+#error "'int' neither 32bit nor 64 bit, I don't know how to handle it."
+#endif
+
+#define FFI_ALIGN(v, a)  (((((size_t) (v))-1) | ((a)-1))+1)
+
+#pragma region "FFI Helpers"
```
saghul, line 81: Never used these before. Do they just help with some editor?

src/js/polyfills/console.js
```diff
@@ -241,3 +241,10 @@ Object.defineProperty(window, 'console', {
     writable: true,
     value: new Console()
 });
+
+Object.defineProperty(window, 'util', {
```
saghul, line 245: Why is this here? If we need any specific utils we should expose them in the `tjs` namespace.

src/js/tjs/ffi.js
```diff
@@ -0,0 +1,716 @@
...
+		this._cif = new ffiInt.FfiCif(rtype.ffiType ?? rtype, ...argtypes.map(t => t.ffiType ?? t));
+		this._closure = new ffiInt.FfiClosure(this._cif, this._func);
+	}
+	get addr(){
+		return this._closure.addr;
+	}
+}
+
+// export only for test, not for user
+export function _parseCProto(header){
```
saghul, line 412: Does this rely on anything not available to userland code? If not, can we move it to some ffiutils.js helper file for the tests?

src/vm.c
```diff
@@ -72,6 +72,7 @@ static void tjs__bootstrap_core(JSContext *ctx, JSValue ns) {
     tjs__mod_wasm_init(ctx, ns);
     tjs__mod_worker_init(ctx, ns);
     tjs__mod_xhr_init(ctx, ns);
+    tjs__mod_ffi_init(ctx, ns);
```
saghul, line 75: Can you please add it alphabetically sorted?

src/utils.h
```diff
@@ -116,4 +116,7 @@ extern size_t tjs_signal_map_count;
 const char *tjs_getsig(int sig);
 int tjs_getsignum(const char *sig_str);
 
+#define THROW_ARG_ERR(ctx, argno, expected) JS_ThrowTypeError(ctx, "expected argument %d to be %s", argno+1, expected)
```
saghul, line 119: add `TJS_` prefix please.

saghul (changes requested):
Left some comments, amazing work!

I'd be happy landing it once those comments (specially the build system one) have been addressed, and then  iterating on it once it's in.

Thank you!

.github/workflows/ci.yml
lal12: To test the CI in my fork, wasn't intended to land in the PR.

Dockerfile
lal12: Yeah probably some leftovers from before I added the embedded ffi.

src/ffi.c
lal12: Didn't used them either before, usually I split into more c files. But considering how everything else was setup I kept it to one file, but it got quite long. And yes it gives you folding e.g. in vscode.

saghul: 👍

src/js/polyfills/console.js
lal12: Debug leftover. Though on second thought having all of util included makes sense IMHO, but another issue…

src/ffi.c
lal12: There shouldn't, but I guess adding a fallback is a good idea.

src/js/tjs/ffi.js
lal12: Not exactly sure how you meant it, look at the new commits.

lal12:
Btw. don't know how you handle this usually, but I would recommend doing a squash merge, cause the PR commit history is a bit messy.

saghul:
> Btw. don't know how you handle this usually, but I would recommend doing a squash merge, cause the PR commit history is a bit messy.

Yep, squash-merge is how I usually do it.

saghul:
Looks like FFI_BAD_ARGTYPE was added not that long ago and the windows package is too old, we might need to ifdef it out. https://github.com/libffi/libffi/commit/205cf01b57972fdc8c090fc79192b464dc43fc0d

lal12:
> Looks like FFI_BAD_ARGTYPE was added not that long ago and the windows package is too old, we might need to ifdef it out. [libffi/libffi@205cf01](https://github.com/libffi/libffi/commit/205cf01b57972fdc8c090fc79192b464dc43fc0d)

Yes, had the same issue in my cross toolchain.

saghul:
Looks like `fabsf` might not be defined in mingw?

```
 test-ffi.js                              FAIL
status:
{ exit_status: 1, term_signal: null }
stderr:
InternalError: uv_dlopen failed: The specified procedure could not be found.
```

lal12:
> Looks like `fabsf` might not be defined in mingw?
> 
> ```
>  test-ffi.js                              FAIL
> status:
> { exit_status: 1, term_signal: null }
> stderr:
> InternalError: uv_dlopen failed: The specified procedure could not be found.
> ```

Haven't looked at it in detail. Though I guess when dlopen failed it is rather that he cannot find the library, otherwise dlsym should have failed. Will give it a look on Monday.

lal12:
Ok, you were right. fabsf apparently is not available in windows or at least this specific build environment. Mingw code has it though... The same is true for opendir. Maybe a different library needs to be included on windows. Well I skip the faulting tests on windows. Using the libc with ffi is platform dependent anyway. And if using libffi works on one platform it will work on the others or it is a bug in libffi.

saghul (approved):
Let's get it!

PR #302, opened 2022-07-25
Added PosixSocket

lal12:
I've added a class `PosixSocket`, which provides "raw" access to most of the posix socket api.

It uses uv_poll as an event handler. It isn't available for windows.

It is quite useful for stuff like unix, raw/packet and netlink sockets. 

The [same question](https://github.com/saghul/txiki.js/pull/299#issuecomment-1192998129) regarding npm/esbuild in the build process apllies.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,8 @@ message(STATUS "Building with ${CMAKE_C_COMPILER_ID} ${CMAKE_C_COMPILER_VERSION}
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -g")
-set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -ggdb -O0 -fno-omit-frame-pointer")
+set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -DDEBUG -ggdb -O0 -fno-omit-frame-pointer")
+
 
 message(STATUS "C flags:         ${CMAKE_C_FLAGS}")
 message(STATUS "C Debug flags:   ${CMAKE_C_FLAGS_DEBUG}")
@@ -44,27 +45,6 @@ endmacro()
 
 cpr_option(BUILD_WITH_ASAN "If ON, build with the address sanitizer enabled" OFF)
 
-configure_file(
-        ${CMAKE_SOURCE_DIR}/src/js/bundle.js
-        ${CMAKE_CURRENT_BINARY_DIR}/bundle.js
-        COPYONLY)
-configure_file(
-        ${CMAKE_SOURCE_DIR}/src/js/eval-stdin.js
-        ${CMAKE_CURRENT_BINARY_DIR}/eval-stdin.js
-        COPYONLY)
-configure_file(
-        ${CMAKE_SOURCE_DIR}/src/js/repl.js
-        ${CMAKE_CURRENT_BINARY_DIR}/repl.js
-        COPYONLY)
-configure_file(
-        ${CMAKE_SOURCE_DIR}/src/js/std.js
-        ${CMAKE_CURRENT_BINARY_DIR}/std.js
-        COPYONLY)
-configure_file(
-        ${CMAKE_SOURCE_DIR}/src/js/worker-bootstrap.js
-        ${CMAKE_CURRENT_BINARY_DIR}/worker-bootstrap.js
-        COPYONLY)
-
 add_subdirectory(deps/quickjs EXCLUDE_FROM_ALL)
 
 option(libuv_buildtests "" OFF)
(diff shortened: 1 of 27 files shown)


CMakeLists.txt
```diff
@@ -75,6 +55,32 @@ add_subdirectory(deps/wasm3 EXCLUDE_FROM_ALL)
...
+set(INCLUDED_JS_FILES_PATH "")
+foreach(jsfile ${INCLUDED_JS_FILES})
+    configure_file(
+        ${CMAKE_SOURCE_DIR}/src/js/${jsfile}
+        ${CMAKE_CURRENT_BINARY_DIR}/${jsfile}
+        COPYONLY)
+	list(APPEND INCLUDED_JS_FILES_PATH ${CMAKE_CURRENT_BINARY_DIR}/${jsfile})
+endforeach()
+
+add_custom_command(
```
saghul, line 70: Same commend as the other PR, let's discuss this as a separate effort please.

package.json
```diff
@@ -14,6 +14,7 @@
   "license": "MIT",
   "dependencies": {
     "abortcontroller-polyfill": "^1.7.3",
+    "buffer": "^6.0.3",
```
saghul, line 17: I really don't want to depend on this. Vanila JS shoula already have everything we need. What are you missing?

src/js/tjs/posix-socket.js
```diff
@@ -0,0 +1,162 @@
...
+			}
+		}
+		this._psock.poll(mask, this._handleEvent);
+	}
+
+	stopPoll(){
+		this._psock.pollStop();
+	}
+
+	static ip4ToBuf(ipstr){
```
saghul, line 129: I think we have a similar function in the C layer, why do you need this?

src/posix-socket.c
```diff
@@ -0,0 +1,612 @@
...
+    CHECK_ARG_RET(ctx, s, -1, tjs_sock_class.class_name);
+    if(s->closed){
+        return JS_ThrowInternalError(ctx, "Socket closed");
+    }
+
+    int level, optname;
+    CHECK_ARG_RET(ctx, !JS_ToUint32(ctx, &level, argv[0]), 0, "positive integer");
+    CHECK_ARG_RET(ctx, !JS_ToUint32(ctx, &optname, argv[1]), 1, "positive integer");
+    
+    socklen_t optlen = 50; // TODO: adjust value
```
saghul, line 194: what options are so large?

src/posix-socket.c
```diff
@@ -0,0 +1,612 @@
...
+
+
+static void tjs_sock_uv_poll_cb(uv_poll_t* handle, int status, int events){
+	tjs_sock_t* s = uv_handle_get_data((uv_handle_t*)handle);
+	JSValueConst args[] = {
+		JS_NewInt32(s->jsctx, status),
+		JS_NewInt32(s->jsctx, events)
+	};
+	s->in_cb = true;
+	JSValue ret = JS_Call(s->jsctx, s->callback, s->this, countof(args), args);
```
saghul, line 451: I think you need to dup the callback here to avoid problems if you call close in the callback.

saghul (commented):
Interesting stuff!

Generally speaking I've tried to stay away from custom stuff not covered by libuv. This is somewhere in between, because libuv does have uv_poll_t, but needs the application to "help".

What is your use-case? Non TCP non UDP sockets? Do you have an actual use for it or is it more like "I'd like to have this in case i need it someday" kind of thing?

Just trying to measure, since this is another API that will need to be maintained 😅

lal12:
> What is your use-case? Non TCP non UDP sockets? Do you have an actual use for it or is it more like "I'd like to have this in case i need it someday" kind of thing?

No, I definitely need and use it. I have several applications where I use raw/packet sockets (e.g. DHCP packet handling) and unix stream socket to interact with a system bus and unix dgram socket for systemd notify e.g.. 

Basically I am replacing Nodejs with quickjs / txiki, because NodeJS is too large / ressource hungry and a pain to build. Also I had limitations with Node because it does not support all types of sockets I need (even with additional packages).

package.json
lal12: The write* (writeUint32LE, writeUint32BE) functions are quite useful. 

But you're right. I wasn't aware that DataView can handle different endianess too.

src/posix-socket.c
lal12: close in the callback shouldn't happen (libuv doesn't like that). 
So I added a check (like in poll_stop).

saghul:
> > What is your use-case? Non TCP non UDP sockets? Do you have an actual use for it or is it more like "I'd like to have this in case i need it someday" kind of thing?
> 
> No, I definitely need and use it. I have several applications where I use raw/packet sockets (e.g. DHCP packet handling) and unix stream socket to interact with a system bus and unix dgram socket for systemd notify e.g..
> 
> Basically I am replacing Nodejs with quickjs / txiki, because NodeJS is too large / ressource hungry and a pain to build. Also I had limitations with Node because it does not support all types of sockets I need (specifically packet sockts and unix dgram socket) (even with additional packages).
> 
> In case of the posix API I would argue there probably won't be any changes in the foreseeable future xD. (Yeah I know there is the quickjs side which might need maintenance).

Gotcha, thanks for explaining!

src/js/tjs/posix-socket.js
lal12: Removed buffer and used tjs_obj2addr

src/posix-socket.c
lal12: The largest one I could find for SOL_SOCKET is `struct sockaddr_storage` for `SO_PEERNAME` which is 128byte. But there could be larger ones for more specific Socket types. E.g. Bluetooth socket or maybe some more specific TCP/IP options getable via getsockopt. So now I just let the user pass an optional size, otherwise I will use 128 as the default.

lal12: FYI: Just have a case where a socket option can get arbitrarily long, `SOL_SOCKET, SO_ATTACH_FILTER` (BPF Packet filter).

lal12:
I've added some helper functions and an example. 

@saghul  Probably you just didn't came around to take a look. Just wanted to be sure that you know, I consider this done and merge-able.

saghul:
> I've added some helper functions and an example.
> 
> @saghul Probably you just didn't came around to take a look. Just wanted to be sure that you know, I consider this done and merge-able.

Ops, I had missed the updates indeed! I'll take another look!

tests/test-posix-socket.js
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	assert.eq(PosixSocket.checksum(new Uint8Array(
+		`45 00 00 99 12 9f 40 00 01 11 00 00 0a 00 00 7a ef ff ff fa`.split(' ').map(x=>parseInt(x, 16))
+	)), 0x416c);
+}
+
+async function run(){
+	if(tjs.platform == 'windows'){
+		// windows (new versions at least) doesn't support posix.
```
brandonros, line 113: ws2_32/winsock2 don't support POSIX-like sockets?

lal12: The Posix API was removed in/after Windows 8, but maybe Sockets are actually still supported...

lal12: So I've tried it. Yes winsock supports the basic socket api. But recvmsg and sendmsg e.g. are not supported.

I am not sure how useful this would be on Windows anyway. For simple TCP/UDP sockets there is the normal tjs socket api. And more advanced stuff (unix, netlink, raw, packet) won't work on windows anyway.

saghul:
Can you please run clang-format? I see a number of style shenanigans. Which reminds me I should add an eslint file for the JS side...

lal12:
Ok I've run clang-format. Though I only comitted it for the posix-socket.c, since a few other headers also changed.

saghul (approved):

PR #303, opened 2022-08-01
Added os uptime and process uptime functions

lal12:
- Adds function `tjs.uptime` to get the system uptime.
- Adds a function `tjs.processUptime` to get the uptime relative to the process start

Reasons for adding this:
- `tjs.uptime` : hrtime is not guaranteed to be the uptime
- `tjs.processUptime`: not really any alternative in the current api. Doing it in JS probably has worse accuracy.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -164,8 +164,6 @@ TJSRuntime *TJS_NewRuntimeWorker(void) {
 }
 
 TJSRuntime *TJS_NewRuntimeInternal(bool is_worker, TJSRunOptions *options) {
-    if(tjs_process_start_uptime == -1) 
-        uv_uptime(&tjs_process_start_uptime);
     TJSRuntime *qrt = calloc(1, sizeof(*qrt));
 
     memcpy(&qrt->options, options, sizeof(*options));
--- a/src/vm.c
+++ b/src/vm.c
@@ -164,6 +164,7 @@ TJSRuntime *TJS_NewRuntimeWorker(void) {
 }
 
 TJSRuntime *TJS_NewRuntimeInternal(bool is_worker, TJSRunOptions *options) {
+    uv_uptime(&tjs_process_start_uptime);
     TJSRuntime *qrt = calloc(1, sizeof(*qrt));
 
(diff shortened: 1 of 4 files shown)


saghul:
What do you neeed `processUptime` for?

src/vm.c
```diff
@@ -164,6 +164,8 @@ TJSRuntime *TJS_NewRuntimeWorker(void) {
 }
 
 TJSRuntime *TJS_NewRuntimeInternal(bool is_worker, TJSRunOptions *options) {
+    if(tjs_process_start_uptime == -1)
```
saghul, line 167: Maybe better use a uv_once for this, since there could be multiple threads started with different runtimes.

lal12:
I used ProcessUptime mostly for debugging. I guess (new Performance()).timeOrigin could be used instead, however this doesn't really seem to be guaranteed (by the standard) being the process uptime. Probably cpu times would be more useful in that case too, however I haven't seen this to be exposed for the current process.

Everything else could be substituted by just uptime, I only used processUptime for semantic reasons there.

Though I am not really "passionate" about adding processUptime (or runtimeUptime). I only really need system uptime.

lal12:
So removing that or changing it to "runtime uptime" is your decision, I don't care.

saghul:
Do you care about the semantics or about the specific implementation? If you care about the latter, then `performance.timeOrigin` is a good choice because it references the start of the runtime (when the globals got assigned) and it's value is guaranteed to be monotonically increasing and any new perf metrics won't go back in time and will give you an accurate measurement since the start of the runtime.

Since every new runtime has a new JS heap, they'll have different time origins.

lal12:
I mostly was worried about timeOrigin not being defined as a specific unit by the standard, but for debugging I guess I can rely on the specific TJS behavior anyway.

And the for the production part it really was only semantics, e.g. for workers and IPCs. But using the system uptime should work just as well. I will remove the processUptime.

saghul (approved):

saghul:
Thanks!

PR #311, opened 2022-09-15
js: change to UTF-8-only text-encoding package to reduce bundle size

samuelbrian:
The text-encoding package was automatically including all the encoding indexes, which was about 600kb of the previously 1100kb bundle.js. We only need UTF-8 for ArrayBuffer encoding/decoding, so removing the other encodings saves us tjs footprint size and process start time. The bundle.js is now 500kb.


saghul:
Thank you!

PR #315, opened 2022-10-19
FFI: fixed Pointer.deref() and DlSymbol.addr

samuelbrian:
Hi @saghul. I found and fixed some problems with FFI.


samuelbrian:
Updated to pass linter.

saghul:
Thank you!

PR #321, opened 2022-11-04
TextDecoder pollyfill improvement and TypeScript declaration fix for DirHandle

samuelbrian:
Sorry I missed your release with these @saghul.


saghul:
> Sorry I missed your release with these @saghul.

No worries, numbers are cheap! 😅

PR #322, opened 2022-11-04
os: add chdir() from libuv

samuelbrian:
I could use chdir() via FFI, but since it's cross-platform through libuv I figured it's a pretty easy inclusion into tjs.

### DIFF
--- a/docs/txikijs.d.ts
+++ b/docs/txikijs.d.ts
@@ -176,6 +176,11 @@ declare namespace tjs {
     function exit(code: number): void;
 
     /**
+     * Changes the current working directory.
+     */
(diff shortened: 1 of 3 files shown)


saghul:
Makes sense, thanks! The path module we currently ship is posix only though, so the path compare fails. I think if you use `tjs.realpath()` on the result it will normalize it properly.

Going forward we probably need to embed a proper Windows path module.

samuelbrian:
Thanks for the suggestion @saghul. All passing now.

saghul (approved):

saghul:
Thank you!

PR #357, opened 2023-01-23
crypto,uuid: use a native UUIDv4 implementation

saghul:


saghul:
I still need to fixup the uuid4 library:

- make it use libuv's random byte generator
- make the seed thread local / a context struct?

bnoordhuis:
I don't know what security properties you're aiming for but that library uses xorshift128+, which is not a CSPRNG.

Maybe easiest is to obtain 16 bytes of randomness from libuv, patch in the 6 uuidv4 marker bits, and format it.

bnoordhuis:
I suppose I should put my money where my mouth is. :-)

What I mean is this:
```c
// usage:
//  char s[37];
//  int err = uuidv4(&s);
int uuidv4(char (*s)[37]) {
  unsigned char u[16];
  int err;

  err = uv_random(NULL, NULL, u, sizeof(u), 0, NULL);
  if (err)
    return err;

  u[6] &= 15;
  u[6] |= 64; // '4x'

  u[8] &= 63;
  u[8] |= 128; // 0b10xxxxxx

  snprintf(*s, sizeof(*s),
           "%02x%02x%02x%02x-%02x%02x-%02x%02x-"
           "%02x%02x-%02x%02x%02x%02x%02x%02x",
           u[0], u[1], u[2], u[3], u[4], u[5], u[6], u[7],
           u[8], u[9], u[10], u[11], u[12], u[13], u[14], u[15]);

  return 0;
}
```

saghul:
Ha, even simpler! I ended up replacing the seed material with uv_random. But I like yours better :-)

PR #378, opened 2023-02-23
Add build-essential to avoid build error

guest271314:
make

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.

After installing build-essential make builds successfully.


saghul:
Cheers!

PR #391, opened 2023-09-15
added @txikijs/types

lal12:
This seems to work so far.

Unfortunately tsdoc now seems to ignore the `@module` directive as well as the the file header comments.

Publishing it via github actions would also be nice, however since it requires 2FA I don't think it would be useful.


saghul:
Thank you! I'll take a look!

Have you checked the docs page still works after these changes?

saghul:
Thank you! I'll take a look!

Have you checked the docs page still works after these changes?

lal12:
Yes it does work, but as mentioned above unfortunately not quite as before.

That's why it still a draft, hoped you might have an idea, otherwise I will look into that next week

lal12:
Hmm tried quite a lot, but I didn't got it to work. Either TS types don't properly work or typedoc names modules as e.g. assert instead of tjs:assert.

Have left a question in the typedoc repo…

saghul:
Quick update: you might have seen some recent activity in the repo :-) I'm mostly done with the changes I wanted to have ready for the next release, I'll be looking into this PR soon!

saghul:
Hum, I took a look and in addition to the module name the comments there are also missing, which is not great.

I don't think I'll have time to dig into typedoc anytime soon.

Do you have any suggestions on the way forward? I like the idea of having this PR in, but it breaking the (small amount) of docs the project has is not great...

lal12:
It's currently tracked as a bug here: https://github.com/TypeStrong/typedoc/issues/2401

lal12:
With the updated typedoc the resulting doc is identical to the one before.

lal12:
Had some issues with package-lock, should now be fixed.

saghul:
This looks great thanks for working on it! ❤️

lal12:
This still needs to be published. I already created the @txikijs org in npm. I've sent you an invite.

Unfortunately I accidentally burned the 23.1.0 version during testing. I would just use 23.1.1 or do you have another proposal?

saghul:
We should also automate it with a GH actions perhaps that runs every time a release is created.

lal12:
I considered that initially. However this would require disabling 2FA. Usually not a good idea due to supply chain attacks. 

And with 2FA there is a quirky way through GH actions (https://github.com/marketplace/actions/npm-otp-publish), but I would consider it easier to just publish by hand than waiting for some GH Action and then paste my OTP somewhere.

But maybe you would prefer it that way?

saghul:
Isn't it possible to get a token that bypasses 2FA? I seem to recall using those for other projects.

lal12:
Seems you're right: 
image

PR #409, opened 2023-10-03
fix: mingw toExponential toPrecision toFixed

zeromake:


saghul:
I'll admit I'm out of my depth here.

Can you please elaborate on the rationale behind the fix?

zeromake:
@saghul 
I changed the implementation and this 1e-6 can't handle all cases.
the new scenario is to find that there is 5 to 6 only on mingw
```js
(25).toExponential(0) == '3e+1' // 25 + 1
(2.5).toPrecision(1) == '3' // 2.5 + 0.1
```

zeromake:
I saw that there are more situations, I will adjust it locally and then turn it on

PR #381, opened 2023-03-01
autoconf, textinfo dependencies Debian/Ubuntu

guest271314:
autoconf for autoreconf, textinfo for makeinfo dependencies Debian/Ubuntu

PR #413, opened 2023-10-09
Add libtool to install dependencies

topelinux:
without libtool install the build is failed with following error message
`Makefile.am:38: error: Libtool library used but 'LIBTOOL' is undefined`


saghul:
Should be covered by https://github.com/saghul/txiki.js/pull/381

PR #417, opened 2023-10-28
Added QuickJS Debugger Support

tritao:
Integrated quickjs debugger support from https://github.com/koush/vscode-quickjs-debug

Adds the ability to debug javascript with Visual Studio Code (using a quickjs debugger extension).

QuickJS Debugger supports stepping, continue, breakpoints, evaluation, and variable access.

QuickJS Debugger support is not compiled in by default. Given tight integration within the quickjs opcode dispatcher and unsure of performance implications.

To enable support
```
mkdir build; cd build; cmake .. -DCONFIG_QUICKJS_DEBUGGER=ON; make
```

Rebased from https://github.com/saghul/txiki.js/pull/184, original credit goes to @johnrobinsn.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,7 @@ endmacro()
 
 cpr_option(BUILD_WITH_ASAN "If ON, build with the address sanitizer enabled" OFF)
 cpr_option(USE_EXTERNAL_FFI "Specify to use external ffi dependency" OFF)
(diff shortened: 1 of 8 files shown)


saghul:
Let's please move this over to https://github.com/quickjs-ng/quickjs

@bnoordhuis and myself are reigniting the QuickJS development there.

I reckon it would be best to have a discussion first.

PR #428, opened 2023-11-13
file: embed polyfill for blob and file

jimmywarting:
This is a kind of better blob implementation cuz it can references other blobs and dose not have to hold all data in memory. 
it's a way better implementation for later being able to create blob's that are backed up by the disk.

### DIFF
--- a/src/js/polyfills/blob.js
+++ b/src/js/polyfills/blob.js
@@ -1,4 +1,240 @@
-import { Blob } from 'blob-polyfill';
+// 64 KiB (same size chrome slice theirs blob into Uint8array's)
+const POOL_SIZE = 65536;
+const { isView } = ArrayBuffer;
+
+/**
+ * @param {(Blob | Uint8Array)[]} parts
+ * @returns {AsyncIterableIterator<Uint8Array>}
+ */
+async function * toIterator (parts) {
+    for (const part of parts) {
+        if (isView(part)) {
+            let position = part.byteOffset;
+            const end = part.byteOffset + part.byteLength;
+
+            while (position !== end) {
+                const size = Math.min(end - position, POOL_SIZE);
+                const chunk = part.buffer.slice(position, position + size);
+
+                position += chunk.byteLength;
+                yield new Uint8Array(chunk);
+            }
+        } else {
+            yield * part.stream();
+        }
+    }
+}
+
+class Blob {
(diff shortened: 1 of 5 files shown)


saghul:
Thank you! Can you regenerate the C file please?

jimmywarting:
Plz Tell me how. This is my first contribution.

saghul:
Ssure, just run `make js`

jimmywarting:
Thx, Done!

saghul:
Looks like a test is failing, can you check?

jimmywarting:
A blob where constructed with invalid arguments in `test-ws-send.js`
```diff
- new Blob(arrayBuffer)
+ new Blob([ arrayBuffer ])
```

src/js/polyfills/blob.js
```diff
@@ -1,4 +1,240 @@
...
+     * @return {Promise<string>}
+     */
+    async text () {
+        // More optimized than using this.arrayBuffer()
+        // that requires twice as much ram
+        const decoder = new TextDecoder();
+        let str = '';
+
+        for await (const part of this.stream()) {
+            str += decoder.decode(part, { stream: true });
```
saghul, line 125: We don't currently support the stream parameter, and isn't it deprecated?

jimmywarting: um...? i'm confused. 
i did not find anything on [mdn](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/decode#specifications), the [spec](https://encoding.spec.whatwg.org/#ref-for-dom-textdecoder-decode), or on [github](https://github.com/whatwg/encoding/issues) about it being deprecated. 

i also tried it out in txiki and streaming dose seem to be supported?

image

as to compared to:

```js
decoder.decode(new Uint8Array([226, 153]));
// "��"
decoder.decode(new Uint8Array([165]));
// "�"
```

saghul: You are right, not sure what I confused this with 😅 It it supported indeed, in the polyfill we use. This is ready to go then!

saghul:
Thanks a lot!

saghul:
I was thinking about using your FormData libary to polyfill FormData, what do you think?

jimmywarting:
Sure. Feel free to copy as much or as little as you want

PR #429, opened 2023-11-14
impl-domexception

jimmywarting:
Wanted to have the most basic low level stuff shipped first and formost so other web apis that depends on it can start using it and working on those "Minimum Common Web Platform API"

DOMException is surprisingly used by so many stuff. 
atob, postMessage, web streams, file system access (whatwg/fs), AbortController and so many more

So it needs to come first before work on atob etc can be worked on...


saghul (approved):

saghul:
Nice!

PR #439, opened 2023-12-11
web-api: add navigator polyfill

saghul:


src/js/polyfills/navigator.js
guest271314, line 1: Bun
```
'Bun/1.0.15'
```

Deno

```
"Deno/1.38.5"
```

Node.js

```
'Node.js/22'
```

Other runtimes are including `"/<version>"` in the value, too.

saghul: The spec recommends against it.

Once the run time is detected tjs.version has the version.

Issue #453, opened 2024-02-19
Build problems on Void Linux


ItsTheWhale:
# Problem

Attempted to build the software (commit `67f7059`) but having CMake errors.

# Environment

Void Linux with Linux 6.6_1 on x86_64, with installed dependencies `base-devel` (20181003_2), `libcurl-devel` (8.6.0_1), `cmake` (3.27.6_1), `autoconf` (2.71_1), `texinfo` (7.0.3_1), `libtool` (2.4.7_5).

# Expected Result

Successful build with no errors.

# Actual Result

CMake error.
```
$ git clone --recursive https://github.com/saghul/txiki.js --shallow-submodules
$ cd txiki.js
$ make
cmake -B build -DCMAKE_BUILD_TYPE=Release
-- The C compiler identification is GNU 13.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Building in Release mode
-- Building with GNU 13.2.0 on Linux-6.6.16_1
--   BUILD_WITH_ASAN: OFF
--   USE_EXTERNAL_FFI: OFF
-- Building in Release mode
-- Building with GNU 13.2.0 on Linux-6.6.16_1
-- Performing Test COMPILER_SUPPORTS_Wall
-- Performing Test COMPILER_SUPPORTS_Wall - Success
-- Performing Test COMPILER_SUPPORTS_Werror
-- Performing Test COMPILER_SUPPORTS_Werror - Success
-- Performing Test COMPILER_SUPPORTS_Wextra
-- Performing Test COMPILER_SUPPORTS_Wextra - Success
-- Performing Test COMPILER_SUPPORTS_Wnoimplicitfallthrough
-- Performing Test COMPILER_SUPPORTS_Wnoimplicitfallthrough - Success
-- Performing Test COMPILER_SUPPORTS_Wnosigncompare
-- Performing Test COMPILER_SUPPORTS_Wnosigncompare - Success
-- Performing Test COMPILER_SUPPORTS_Wnomissingfieldinitializers
-- Performing Test COMPILER_SUPPORTS_Wnomissingfieldinitializers - Success
-- Performing Test COMPILER_SUPPORTS_Wnounusedparameter
-- Performing Test COMPILER_SUPPORTS_Wnounusedparameter - Success
[...]

PR #454, opened 2024-02-19
build: fix libffi building to the wrong directory (https://github.com/saghul/txiki.js/issues/453)

ItsTheWhale:
On systems where `gcc --print-multi-os-directory` outputs something that is not `../lib`, Libffi installs to the wrong location (e.g. `lib64`) where CMake cannot find it and errors, breaking the build. This fix disables placing binaries in other places with the `--disable-multi-os-directory` configuration flag, forcing Libffi to install to `lib`.
This fix has no effect on systems where Libffi already installs to `lib`, and therefore does not break anything.


saghul (approved):
Thank you!

PR #456, opened 2024-02-28
Mimalloc

saghul:


saghul:
Ha, this surfaced some real allocation problems!

saghul:
@lal12 Looks like the ffi code has an out of bounds memory write, in computing the size of the return type AFAIS. Does anything stand out to you?

PR #483, opened 2024-04-11
ci: add arm64 macOS to the testing matrix

saghul:


saghul:
@lal12 The sprintf test fails on arm64 macOS. Got any idea why that would be? I can repro locally too.

lal12:
Or disabling libffi for this scenario by default in CMAKE.

lal12:
Unfortunately I don't have access to any arm64 MacOS devices.

lal12:
Just tested with the newest libffi release, now it's a SEGFAULT instead. (https://github.com/lal12/txiki.js/actions/runs/8643918438/job/23697940801)

Hard to debug without a bare metal. Tried scaleway to book one online, but they didn't let me book it.

saghul:
> * there is a fork from apple for libffi which apparently works

Can we try this?

saghul:
Added a commit which links with the system provided libffi, but it segfaults 🙃

lal12:
> Added a commit which links with the system provided libffi, but it segfaults 🙃

I fiddled around a bit. My suspicion is that all is fine with libffi, just that sprintf and dirent have different definitions on MacOS for arm. If I comment both test parts out it works fine. 

I think moving all tests into our own binary testlib (as one test already is), is more robust (also considering the windows exceptions).

saghul:
> I fiddled around a bit. My suspicion is that all is fine with libffi, just that sprintf and dirent have different definitions on MacOS for arm. If I comment both test parts out it works fine.

I see. THe `sprintf` part is surprising TBH, but I'm no expert 😅  I'll ignore it on macOS + arm for the time being then.



> I think moving all tests into our own binary testlib (as one test already is), is more robust (also considering the windows exceptions).

Ah, the ffi-test-lib.c file. Yeah that makes sense! Feel free to do that change when you've got the time!

lal12:
>I see. THe sprintf part is surprising TBH, but I'm no expert 😅 I'll ignore it on macOS + arm for the time being then.

Yep, you're right. At least sprintf has a wrong call in test-ffi.js. Interesting that it works on any other platform.
But I am gonna move everything to the testlib anyway.

PR #484, opened 2024-04-11
move most of test-ffi to own testlib

lal12:
Move most ffi test to testlib instead of using libc, since libc partially differs depending on platform.
This also fixes tests on windows and macos arm.


saghul:
Wonderful!

PR #485, opened 2024-04-12
console implementation according to standard

lal12:
Implementing console according to standard.

Still a bunch of test cases missing, but otherwise complete.

Any thoughts?


lal12:
E.g. would be nice to expose the createConsole function for customization.

src/bundles/c/stdlib/assert.c
```diff
@@ -361,9 +361,9 @@ const uint8_t tjs__assert[8835] = {
  0x75, 0x72, 0x6e, 0x20, 0x61, 0x20, 0x21, 0x3d,
```
saghul, line 361: (random comment placement).

What did you use to regenerate the code? You only changed a polyfill, so stdlib shouldn't be regenerated I think.

src/js/polyfills/console.js
```diff
@@ -1,258 +1,316 @@
...
+    const _printer = (logLevel, ...args) => {
+        printer(logLevel, groupCount, args);
+    };
 
-// Copyright Joyent, Inc. and other Node contributors. MIT license.
-// Forked from Node's lib/internal/cli_table.js
+    if (!logger) {
+        logger = function Logger(logLevel, args) {
+            if (args.length === 1) {
+                _printer(logLevel, args[0]);
```
saghul, line 18: Why is the single argument not formatted here?

src/js/polyfills/console.js
```diff
@@ -1,258 +1,316 @@
...
+            } else if (args.length > 1) {
+                _printer(logLevel, formatter(...args));
+            }
+        };
+    }
 
-function hasOwnProperty(obj, v) {
-    if (obj === null  || typeof obj === 'undefined') {
-        return false;
+    if (!formatter) {
```
saghul, line 25: can we use defaults instead please?

src/js/polyfills/console.js
```diff
@@ -1,258 +1,316 @@
...
+        formatter = utilFormat;
     }
 
-    return Object.prototype.hasOwnProperty.call(obj, v);
-}
+    if (!inspect) {
```
saghul, line 29: Ditto.

src/js/polyfills/console.js
```diff
@@ -1,258 +1,316 @@
...
 
-    for (let i = 0; i < row.length; i++) {
-        const cell = row[i];
-        const len = countBytes(cell);
-        const needed = (columnWidths[i] - len) / 2;
+            logger('assert', data);
+        },
+        clear() {
+            groupCount = 0;
+            clear();
```
saghul, line 58: This reads weird, can we rename the auxiliary function?

src/js/polyfills/console.js
```diff
@@ -1,258 +1,316 @@
...
-class Console {
-    log(...args) {
-        print(...args);
-    }
+            const cols = [];
 
-    info(...args) {
-        print(...args);
-    }
+            for (let ci=0;ci<rows[0].length;ci++) {
```
saghul, line 127: style: please add spacing

src/js/polyfills/console.js
```diff
@@ -1,258 +1,316 @@
...
+        // Counting
+        count(label = 'default') {
+            let count = countMap.get(label) ?? 0;
+
+            count++;
+            countMap.set(label, count);
+            _printer('count', label + ': ' + count);
+        },
+        countReset(label = 'default') {
+            if (!countMap.delete(label)) {
```
saghul, line 218: When is `count` decreased?

src/js/polyfills/console.js
```diff
@@ -1,258 +1,316 @@
...
+            if ([ 'error', 'trace', 'warn' ].includes(logLevel)) {
+                tjs.stderr.write(encoder.encode((' ').repeat(indent) + msg + '\n'));
+            } else {
+                tjs.stdout.write(encoder.encode((' ').repeat(indent) + msg + '\n'));
+            }
+        },
+    })
+});
+
+Object.defineProperty(window, 'createConsole', {
```
saghul, line 311: Let's attach this to the `tjs` global instead, since it's non-standard.

tests/test-console.js
```diff
@@ -0,0 +1,80 @@
...
+
+	// Test console.time() and console.timeEnd()
+	{ code: 'console.time("timer");\nconsole.timeEnd("timer");', result: 'timer: 0ms\n' },
+
+	// Test console.group() and console.groupEnd()
+	{ code: 'console.group("Group 1");\nconsole.log("Hello from Group 1!");\nconsole.groupEnd();console.log("abc")', result: 'Group 1\n Hello from Group 1!\nabc\n' },
+
+];
+
+(async ()=>{
```
saghul, line 39: No need for the immediate function.

tests/test-console.js
```diff
@@ -0,0 +1,80 @@
...
+		const args = [
+			tjs.exepath,
+			'eval',
+			code
+		];
+		const proc = tjs.spawn(args, { stdout: 'pipe', stderr: 'pipe' });
+		const status = await proc.wait();
+		const buf = new Uint8Array(4096);
+		const nread = await proc.stdout.read(buf);
+		const stdout = new TextDecoder().decode(buf.subarray(0, nread));
```
saghul, line 50: Create and reuse a decoder.

tests/test-console.js
```diff
@@ -0,0 +1,80 @@
...
+		return {code: status.exit_status, stdout, stderr};
+	}
+
+	for (const test of tests) {
+		const jscode = test.code;
+		const result = test.result;
+
+		const {code, stdout, stderr} = await runTest(jscode);
+		if(code !== 0){
+			console.error(`Test failed with code ${code}: ${jscode}`);
```
saghul, line 62: This should actually fail the test.

saghul (changes requested):
Great work! Left some comments.

src/bundles/c/stdlib/assert.c
lal12: `make js`. Yes, you're right. I think I was on an unclean checkout (on another notebook) when force pushing it last night.

src/js/polyfills/console.js
saghul: Sorry, my bad, I misunderstood what the API did 😅

saghul, line 218: According to the spec a reset sets it to 0, it doesn't delete it. Is trhere a practical difference?

src/js/polyfills/console.js
lal12: See spec: https://console.spec.whatwg.org/#logger

saghul: 👍

src/js/polyfills/console.js
lal12: Yep, there is. When it is set to 0 you won't get a warning the next time when calling `.countReset()` again.

Though this is a weird case and nodejs does it the way I do. And I think it much better otherwise you never could clean up a counter and it would leak stuff.

```
> console.count('a')
a: 1
undefined
> console.countReset('a')
undefined
> console.countReset('a')
undefined
> (node:18291) Warning: Count for 'a' does not exist
(Use `node --trace-warnings ...` to show where the warning was created)
```

src/js/polyfills/console.js
lal12: Not sure where exactly you meant. Maybe got lost during force-push. But eslint doesn't complain.

src/js/polyfills/console.js
saghul: Very subtle, let's keep it as is not to have an ever growing map indeed.

lal12:
Added tests from wpt and added some changes to fulfill those and match some nodejs behavior. In some places wpt tests seemed to be wrong, and neither browser nor nodejs acted accordingly (but differently from each other), there I used nodejs behavior.

JS console standards seems to be oddly wild :laughing: 

Diverging from spec are the following things:

- countReset is resetting instead setting to 0. MDN leaves it open, NodeJS does this too, Chrome+FF follow spec, though FF also prints `label: 0`
- added options argument to Logger, to allow for some specific warning handling to better match nodejs behavior

I also copied the code from util package and moved it here to add missing features as defined in console spec. Though alternatively we could extend the alteady existing util patch.

saghul:
This looks great! Are you going to be adding anything else, or should we merge?

lal12:
> This looks great! Are you going to be adding anything else, or should we merge?

Just pushed the types for createConsole. I have nothing else to add. The only thing I was wondering if there is a better way for registering `createConsole` (see `console.js:327`)

saghul:
> The only thing I was wondering if there is a better way for registering `createConsole` (see `console.js:327`)

It looks odd indeed. I'll take a look.

saghul (approved):

saghul:
Excellent work @lal12 !

PR #487, opened 2024-04-13
only initalize native ffi code on ffi import

lal12:
I noticed ffi init was always executed on tjs startup. This is only necessary when actually being used / being imported.

Initially I planned adding some init functions in builtins.c, but since this currently only is useful for ffi, I took the small route.


src/ffi.c
```diff
@@ -1045,10 +1045,14 @@ static JSCFunctionListEntry funcs[] = {
     JS_SetPropertyStr(ctx, obj, #name, name##_jsval)
 #define ADD_ALIAS_TYPE(ctx, obj, alias, oldname) JS_SetPropertyStr(ctx, obj, #alias, JS_DupValue(ctx, oldname##_jsval))
 
-void tjs__mod_ffi_init(JSContext *ctx, JSValue ns) {
-    JSValue ffiobj = JS_NewObject(ctx);
+static JSValue ffiobj;
```
saghul, line 1048: Note we allow threads, by way of Workers. This is not safe.

lal12: Oh ok, wasn't aware of that.

Given that I am not sure about the JS_NewClass(ID) thread safety and thread interaction, so a thread local variable for ffiobj probably wouldn't be enough. Having a mutex here also seems to be bit overkill. Maybe implementing such a solution later on when there are more places where it is useful (e.g. lws) might be preferable.

saghul: Class IDs are thread safe, I fixed that in QuickJS-ng by making that field part of the runtime rather than a global.

saghul (approved):
Nice! Can you rebase pl? I updated QuickJS and had to regenerate the bundles.

saghul (approved):

PR #491, opened 2024-04-16
provide core internals via import meta

lal12:
The other tjs.internal symbols could be resolved similar.


saghul:
At a glance I'm not very convinced this is an improvement... If anything, it makes it easier to access the core, which is not something desirable, in general.

lal12:
Why does it make it easier to access the core? Only internal JS should now be able to access the core?

saghul:
> Why does it make it easier to access the core?

Because it's accessible to any JS module, without having to jump through the hidden symbol hoop.



> Only internal JS should now be able to access the core?

That has always been the case 😅  The public API is the `tjs` global and the `tjs:xxx` modules. The core object is a big part of `tjs`, but there are many things we hide or we add to.

lal12:
Sorry my reply was a bit misleading. Completly hiding the internals was exactly the goal. As far as I understand it and as far as I tried it only the internal JS will be able to access `import.meta.core`, while the user code won't have it (`import.meta.core == undefined` -> `true`).

Though maybe I am overlooking something?

saghul:
Hum, is that because you are not adding it to the meta object for non internal modules?

lal12:
> Hum, is that because you are not adding it to the meta object for non internal modules?

Basically. Usually this would also be tricky since import.meta is redefined for every module, but since the internal js is compiled into a single file it is fine.

lal12:
@saghul just to be clear on the state of this. Do you still looking into it / thinking about it? Or are you waiting for me to finish the draft? Or...

saghul:
TBH I'm not sure. I appreciate the elegance of it, but I worry there is no scape hatch if I need one...

I'd like to leave this in pause a little bit since it has no actual impact and come back to it when something tells us it's either the right thing to pursue or drop.

saghul:
Some time has passed and I am currently happy with the current solution. We can always come back to this PR if we find a good reason to do that.

Thanks for the effort though!

PR #497, opened 2024-04-19
console improvements & bugfix

lal12:
- console bugfix: Symbol keys in objects were not printed
- expose format and inspect util in tjs global API


saghul:
The linter seems unhappy 😅  Good stuff, thanks for making `console` way better!

saghul (approved):

PR #499, opened 2024-04-22
misc: remove no longer used JSValueConst type

saghul:


saghul:
I'll be happy to bring it back if / when we come up with a better replacement! Right now it was just noise and some casting.

PR #502, opened 2024-04-30
core: fix timers typo

topelinux:
fix clearInterval typo


saghul:
🤦

saghul:
Can you regenerate the bytecode? make js and commit the result.

topelinux:
> Can you regenerate the bytecode? make js and commit the result.

done, please review again.

saghul:
Looks like there is a mismatch. You can try again by making sure to use Node 20 or I can help later today!

topelinux:
> Looks like there is a mismatch. You can try again by making sure to use Node 20 or I can help later today!

i have try with node 20.5.1, same result(make js no different with git status), maybe there is something i missed, please help to commit on your side.

saghul:
Created https://github.com/saghul/txiki.js/pull/509 since I cannot push to your repo. The commit author is still you.

PR #510, opened 2024-05-06
fix some ffi types

### LINKED ISSUE #508: Strange type mapping for ffi
By the way, there is also something quite suspicious happening @ https://github.com/saghul/txiki.js/blob/master/src/js/stdlib/ffi/ffi.js

```
    [ types.uint32, [ 'uint32_t', 'signed long long', 'signed long long int', 'long long', 'long long int' ] ],
```
Does not seems right, unless there are strange reasons behind this choice.

_Originally posted by @KaruroChori in https://github.com/saghul/txiki.js/issues/506#issuecomment-2095316697_

lal12:
fixes #508


saghul (approved):

saghul:
Thanks for the quick fix @lal12 !

PR #517, opened 2024-05-14
Allow embedding txiki in other applications

jspngh:
Hi,

First of all, thanks for all the work on txiki.
Currently only a command-line interface is available, but I would like to embed this javascript runtime in my own application.

Is this change something that you would be open to?
It requires only a small adaptation of the CMakeLists.txt and Makefile.

Let me know what you think.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,8 +28,8 @@ set(TJS__VERSION_MINOR 12)
 set(TJS__VERSION_PATCH 0)
 set(TJS__VERSION_SUFFIX "")
 configure_file(
-    "${CMAKE_SOURCE_DIR}/src/version.h.in"
-    "${CMAKE_SOURCE_DIR}/src/version.h"
+    "${CMAKE_CURRENT_SOURCE_DIR}/src/version.h.in"
+    "${CMAKE_CURRENT_SOURCE_DIR}/src/version.h"
 )
 
 macro(cpr_option OPTION_NAME OPTION_TEXT OPTION_DEFAULT)
@@ -68,9 +68,8 @@ endif()
 
 find_package(CURL REQUIRED)
 
-add_executable(tjs
+add_library(tjs STATIC
     src/builtins.c
-    src/cli.c
     src/curl-utils.c
     src/curl-websocket.c
     src/dns.c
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -152,13 +151,22 @@ string(TOLOWER ${CMAKE_SYSTEM_NAME} TJS_PLATFORM)
...
 target_include_directories(tjs PRIVATE ${CURL_INCLUDE_DIRS})
+target_include_directories(tjs PUBLIC src)
 target_link_libraries(tjs qjs uv_a m3 sqlite3 m pthread ${CURL_LIBRARIES})
 
 if (BUILD_WITH_MIMALLOC)
     target_compile_definitions(tjs PRIVATE TJS__HAS_MIMALLOC)
     target_link_libraries(tjs mimalloc-static)
 endif()
 
+add_executable(tjs-cli EXCLUDE_FROM_ALL
```
saghul, line 162: Don't exclude it, it needs to always be built.

Makefile
```diff
@@ -22,7 +22,7 @@ $(BUILD_DIR):
 	cmake -B $(BUILD_DIR) -DCMAKE_BUILD_TYPE=$(BUILDTYPE)
 
 $(TJS): $(BUILD_DIR)
-	cmake --build $(BUILD_DIR) -j $(JOBS)
+	cmake --build $(BUILD_DIR) --target tjs-cli -j $(JOBS)
```
saghul, line 25: You don't need this if you don't exclude it.

CMakeLists.txt
saghul: If you'd like to only build the library, then add some BUILD_ONLY_LIBRARY option please.

CMakeLists.txt
```diff
@@ -152,13 +151,22 @@ string(TOLOWER ${CMAKE_SYSTEM_NAME} TJS_PLATFORM)
...
+    src/cli.c
+)
+target_link_libraries(tjs-cli tjs)
```
saghul, line 165: Don't we need all the linked libraries here instead?

saghul (changes requested):
Alright, let's do this! Left some comments.

CMakeLists.txt
jspngh: For me, it works without specifying all the libraries.
I think we can either:
- leave it as it is now
- add the `PUBLIC` scope for the `target_link_libraries` of the tjs library
- add all the libraries here again

What do you prefer?

saghul: Let's leave it for now then.

jspngh:
> Alright, let's do this! Left some comments.

Should be fixed now, let me know if there's anything else.

saghul (approved):

PR #522, opened 2024-05-18
Replaced `any` with `unknowns` in txikijs.d.ts

KaruroChori:
`unknown` in my opinion better matches the semantic of ts types for this situation. Functionally there is no difference since the code implementing that interface is written in js.

### DIFF
--- a/docs/types/txikijs.d.ts
+++ b/docs/types/txikijs.d.ts
@@ -1048,35 +1048,35 @@ declare global {
     */
     function createConsole(opts: {
         /** function to print messages to somewhere, see https://console.spec.whatwg.org/#printer */
-        printer: (logLevel: string, args: any[], options: ConsolePrinterOptions) => void,
+        printer: (logLevel: string, args: unknown[], options: ConsolePrinterOptions) => void,
         /** function to handle normal log messages, see https://console.spec.whatwg.org/#logger */
-        logger?: (logLevel: string, args: any[], options: ConsolePrinterOptions) => void,
+        logger?: (logLevel: string, args: unknown[], options: ConsolePrinterOptions) => void,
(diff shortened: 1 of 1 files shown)


lal12:
Hmm I don't necessarily agree.
The only real advantage of unknown is when using the unknown value, for a parameter it is pretty much useless. Also from the perspective of the user/caller passing "unknown" doesn't really make sense, but passing "any" value is clear.
Other typings like the ts libs and @types/node also use any.

KaruroChori:
When using `any` you are effectively telling the type system that you are giving up on characterizing it AND any later code should disregard that as well.
On the implementation side, it means you can use those variables in any way you want, and no static check is going to be performed at compile time.
Using `unknown` means there is a lack of information about its type, but at some point the implementation is forced to perform a cast an take responsibility of that in order to perform any meaningful operation. Clearly the library is written in JS, so the two options are behaviorally equivalent, but their semantic is hardly the same.

The fact other libraries are using `any` in place of `unknown` does not make it more or less right to be honest. Historically speaking `unknown` is way more recent as it was introduced by ts 3.0. So it does make sense a lot of the practice around it and the libraries already established were using `any` and are still doing it to this day.
I have the feeling that if we were to compare the respective usage in interfaces between node and a more recent runtime like bun, the difference in usage would be more visible, but I have no data to support that.

lal12:
Yeah I get that / know the difference. I just didn't see how it is useful (even just semantically) for an API typing.

But I didn't really consider that this where callbacks which can be user provided, so yes unknown definitely makes sense here. I looked at deno typings, they also use unknown for such cases, though they use any for API where the user does not have to interact.

saghul (approved):

saghul:
Thanks for taking a look @lal12 !

PR #524, opened 2024-05-19
Module manager POC

KaruroChori:
Proof of concept of the external module system. It works! 
It can be tested together with the demo module at https://github.com/KaruroChori/txiki-module-demo.

You should add a `modules.json` file with something like this inside:
```
{
    "trialmode": "/archive/shared/projects/txiki-module/"
}
```

Fetching and unpacking from the internet is not yet supported, but you are meant to be able to just use the http[s] prefix.

Feratures:
- [  ] Repackaging
- [  ] Tests
- [  ] Examples
- [  ] Benchmarks
- [x] Typescript types 
- [x] JS code
- [X] C code


KaruroChori:
The current implementation is using shell commands to untargz the module if downloaded. 
Considering windows is also a target, this is not ideal. A library should be picked to do this job in a more portable way.

CMakeLists.txt
```diff
@@ -68,8 +81,12 @@ endif()
 
 find_package(CURL REQUIRED)
 
+file(GLOB_RECURSE src_extras "src/extras/**/*.c" "src/extras/**/*.cpp")
+file(GLOB src_bundles_extras "src/bundles/c/extras/*.c")
+
 add_library(tjs STATIC
     src/builtins.c
+    src/cli.c
```
jspngh, line 89: You are re-adding the `cli.c` file, but that contains the main function and shouldn't be part of the library.

KaruroChori: Thanks for the report, it was a regression when I merged some commits from master.

saghul:
I've a bit busy these days, I hope to take a close look at this next week!

KaruroChori:
I finally added support to include external native deps as part of the build step of a custom runtime.  
For example, a lvgl-js module can now ask for lvgl & lvgl_drivers to be cloned alongside the other deps.

As I received no feedback so far, I just implemented everything according to my needs on a separate branch, and back-ported it here. I am in the process of documenting how to use the features suggested in this PR on this [separate repo](https://github.com/KaruroChori/txiki-modules).

Aside from bugfixing, I have no plan for more features.

KaruroChori:
> I've a bit busy these days, I hope to take a close look at this next week!

No problem. Please, let me know if you encounter any issue, as the documentation for this is still spotty at best & the implementation a moving target up until now.

KaruroChori:
@saghul, this PR has not been evaluated, or at least it has received not feedback so far.
I am now planning a further expansion, but forcing compatibility with this branch when it is not even clear if the module feature has a path for integration would make no sense; so I am basing it directly on my divergent fork.
Basically I am shifting the tjs interface around (https://github.com/saghul/txiki.js/discussions/539) to be less redundant and crowded, moving to a semantic versioning scheme, and few other changes on top of what has diverged already.

As a side effect, it means that all node packets I ported over or natively wrote for txiki to be used as either npm dependencies or external modules will have to be updated and they are no longer going to work with this master branch.
If there is interest, the best I can do is to branch them off as they are now for future reference or usage.

saghul:
Now that I got some of the bigger things out of the way I hope to take a look at this PR this week.



> Basically I am shifting the tjs interface around (#539) to be less redundant and crowded, moving to a semantic versioning scheme, and few other changes on top of what has diverged already.

I think it's better to have that discussion before diverging (but feel free to do whatever you want with your project, of course!) because I'm not sure I'll go for it.

KaruroChori:
No worries, I am not really trying to push anyone into doing anything or to affect your vision behind this project :D.  

But my projects have needs on their own; so, my general attitude is contribute to master what is feasible in time and aligned in scope to the main branch.
That being said some degree of forking is mostly inevitable.

Still, I am more than open to discuss them early to keep the amount of divergence to a minimum. 
As for the interface, I think there are some objective flaws. They all have limited severity, but the compounded effect makes the experience a bit frustrating:
- Redundancy of some elements
- Sometimes objects are exposed, sometimes functions returning objects.
- Inconsistent casing
- Limited documentation of most entries
- Lack of versioning guidelines to ensure new releases are compatible with older code (which goes back to the issue I have with versioning)

But this discussion should probably continue in #539 or whatever evolved proposal derived from it.

saghul:
Hey @KaruroChori sorry for the delay. I've been giving this a thought this week. It's a lot of code to bring in on one go, so let's break it down.

Your approach consists of 3 parts:

1) The ability to add native code to the build
2) The ability to add JS code to the build
3) Some scripting to make 1 and 2 easier

I wonder what other options we have, so we can get closer, in a way that doesn't seem to be that involved.

For starters, 3) could live out of the project at least while we sort things out.

For adding native code to the build there are a few options:

- Adding support for shared library modules: this request comes and goes, and I wanted FFI to be the answer to that... but it wouldn't hurt to add.
- Since you can now build txiki.js as a library, add the missing APIs so we turn this around and make a new application which embdeds txiki.js, initializes the engine, registers some extra modules and starts the interpreter.

For adding JS code:

- The C API could allow registering external JS modules the same way it already does with the stdlib.
- Using the ability to create a standalone executable, bundle the code that way.

WDYT? Would any of these help you accomplish your goals?

KaruroChori:
Hopefully I will be able to properly reply over the weekend. I have to catch up with the latest changes in this repo first as you have been very active :D.

KaruroChori:
> Your approach consists of 3 parts ...

As far as this repository would be concerned yes.  
The specifications for the `module.json` file and how to structure a module are not relevant: point 3 serves as glue to keep `txiki.js` agnostic in this respect.

> For starters, 3) could live out of the project at least while we sort things out.

While possible, it is just a single file implementing a CLI which must be manually invoked to do anything.
But I have nothing against addressing the rest of the proposal first.  
As you suggested, that script is mostly arbitrary and it implements external specifications, so it can live outside.

> Adding support for shared library modules: this request comes and goes, and I wanted FFI to be the answer to that... but it wouldn't hurt to add.
> Since you can now build txiki.js as a library, add the missing APIs so we turn this around and make a new application which embdeds txiki.js, initializes the engine, registers some extra modules and starts the interpreter.

Both these options are valid but different from what I am trying to achieve.  
With FFI one has no access to the very low level quickjs and txiki specific code, as it is not being exposed.  
So it would not be ok to extend the runtime, only to expose self-contained & high performance modules.  

Using txiki as a library in this sense is more flexible, and might work in several cases. However, there is additional complexity to embed the project, and I would personally find it a bit of an overkill to just add some javascript modules as part of the runtime, even more so if I am not authoring them.

The solution I proposed only requires minimal changes to the current code to add some injection points for custom files. This allows to integrate our arbitrary C/C++ and JS code from other sources and keep it outside the git branch. 

> The C API could allow registering external JS modules the same way it already does with the stdlib.

This does not really solve the issue for hybrid modules.

> Using the ability to create a standalone executable, bundle the code that way.

This does also not solve the issue of hybrid modules on its own, and it requires each executable to bundle any part that we want extra in the runtime.
Yet, the main reason to add features to the runtime is so that scripts and executables can be smaller when distributed. The cache layer I discussed in an other proposal might offer a hybrid approach in this respect, but unlike external modules it has not been implemented.

KaruroChori:
External references to this features and its implementation:
https://github.com/KaruroChori/txiki-modules
https://github.com/KaruroChori/demo-txiki-module

PR #526, opened 2024-05-27
improved posix socket api

### LINKED ISSUE #513: Poll with SOCK_STREAM sockets
While ~`socket.pool`~  `socket.poll` works in conjunction with SOCK_DGRAM sockets, I have not been able to make it work in the context of stream connections (blocking listening / accepting / reading works, I tested it).

Do you know if the relevant functionality is implemented?
I checked in the testsuite but there is only one where example and SOCK_DGRAM is used.
I checked the library code and there is not much screaming "I don't work with sockets" but I am not familiar with libuv, so I cannot tell for sure.

lal12:
- always return a full js instance of posix socket and not internal handle, fixes #513
- always get info from socket handle


saghul:
Some test failures, but the approach looks good!

saghul:
Great work!

PR #528, opened 2024-05-31
Ffi parsing comment support

lal12:
- FFI C prototype parser:
  - Added support for comments
  - Bugfix: void functions where not parsed


lal12:
windows CI seems to be broken?

PR #530, opened 2024-05-31
add AF_UNIX to defines in posix-socket

### LINKED ISSUE #512: AF_UNIX  missing in namespace tjs
Unlike other socket related constants, this is missing.
In general it seems like the library is only focusing on the INET family.

Still, Unix Domain Sockets are working as expected once properly configured, so there is no specific reason to not support them I think.

Also, it would be nice to have js helpers to construct the addr structs.
At the moment we are forced to manually construct the Uint8Array, which works but it is a bit inconvenient.
In case, I can do it.

lal12:
Fixes #512

### DIFF
--- a/docs/types/txikijs.d.ts
+++ b/docs/types/txikijs.d.ts
@@ -954,8 +954,12 @@ declare global {
             static readonly defines: {
                 AF_INET: number,
                 AF_INET6: number,
+                /** only on platforms supporting AF_NETLINK */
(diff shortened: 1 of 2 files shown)


saghul:
Something has changed in msys2, but unrelated to this PR: https://github.com/saghul/txiki.js/pull/531

saghul (approved):

Issue #511, opened 2024-05-06
No support for sql transactions in tjs:sqlite


KaruroChori:
Most query builders and ORM libraries assume there is support for transactions on the database driver.  
From what I can tell this is missing from the custom integration of txiki.

Such change would meaningfully improve the ecosystem of libraries which can be made to work with this runtime.

saghul:
Absolutely! I strated with the bare minimum, with the intention to complete things at a later stage.

Do you have an API in mind we can go over? I got heavily inspired by Bun's when made the txiki.js module.

KaruroChori:
Both better-sqlite3 & bun provide support for transactions, so doing something more or less in line with their implementation would be best in my opinion.
Mostly because the process of integrating dirzzle/kysely/prisma/whatever at that point is just a bit more effort than copy&paste :D.

saghul:
Got it 👍  No ETA at the moment, but it's certainly close in the roadmap.

KaruroChori:
Btw, a lot of libraries I tried were not compatible because of https://github.com/bellard/quickjs/issues/261 which has been just merged to quickjs. What is the general process for patches/fixes in this repo? quickjs <-> quickjs-ng -> txiki.js?

saghul:
I'll port the fox to NG, then bump it here.

PR #536, opened 2024-06-06
sqlite: add transactions support

saghul:
Fixes: https://github.com/saghul/txiki.js/issues/511


KaruroChori:
Great! Fingers crossed, this was the only issue preventing me from making libraries like Drizzle compatible with txiki.

PR #537, opened 2024-06-06
Cherrypick gc features from stable-gluegunfw

KaruroChori:
Restricted version of https://github.com/saghul/txiki.js/pull/520, waiting for upstream support of the threshold getter.
I will keep it as draft up to when the quickjs submodule is updated to a commit supporting the getter.


KaruroChori:
By the way, without callbacks there is no good way to write tests for this feature. I will add them at some point if/when they are merged as well.

src/sys.c
```diff
@@ -156,16 +170,28 @@ static const JSCFunctionListEntry tjs_sys_funcs[] = {
...
     JSValue versions = JS_NewObjectProto(ctx, JS_NULL);
     JS_DefinePropertyValueStr(ctx, versions, "quickjs", JS_NewString(ctx, JS_GetVersion()), JS_PROP_C_W_E);
     JS_DefinePropertyValueStr(ctx, versions, "tjs", JS_NewString(ctx, tjs_version()), JS_PROP_C_W_E);
     JS_DefinePropertyValueStr(ctx, versions, "uv", JS_NewString(ctx, uv_version_string()), JS_PROP_C_W_E);
     JS_DefinePropertyValueStr(ctx, versions, "curl", JS_NewString(ctx, curl_version()), JS_PROP_C_W_E);
     JS_DefinePropertyValueStr(ctx, versions, "wasm3", JS_NewString(ctx, M3_VERSION), JS_PROP_C_W_E);
+
+    JSValue gc = JS_NewObjectProto(ctx, JS_NULL);
+    JS_SetPropertyFunctionList(ctx, gc, tjs_gc_funcs, countof(tjs_gc_funcs));
+    JS_DefinePropertyValueStr(ctx, ns, "$gc", gc, 0);
```
saghul, line 193: Please call this `_gc` so it doesn't get auto-exported to the `tjs` global.

KaruroChori: From my tests because of the 0 in JS_DefinePropertyValueStr it is not exposed.
I can change it but I would like to understand the mechanism.
Is there an automatic operation I missed which is hiding symbols starting with `_`?

saghul: Here is where `_` prefixed things are hidden: https://github.com/saghul/txiki.js/blob/1d936302faa9c6342c964aa9b0bbeb547faf0b77/src/js/core/index.js#L56

KaruroChori: Thank you for the handle, I missed it entirely.

KaruroChori: Done

saghul:
Alright, master is ready for this PR now, I also merged the GC threshold getter API. If you rebase and get the build green we are good to go!

KaruroChori:
In theory it is ok. I have no idea how to make CI/Lint happy, codegen is just because my environment is different I guess, and the alpine test failing is just randomly happening from time to time.

saghul:
You can make the linter happy by running the formatter yourself: npm run format

Codegen relies on the node version, make sure you are using 20.

I can take care of those myself if you can't.

KaruroChori:
I fixed the logic since it was all a bit wonky.

> You can make the linter happy by running the formatter yourself: npm run format

`npm run format` does nothing.  
On this repo I usually run the lint command which is specified in scripts (forcing the fix flag).  

I don't have a node 20 instance on this machine, if you could do that it would be better.

saghul (approved):

saghul:
Ah I meant `make format` sorry. I updated the PR, everything should be green now.

KaruroChori:
Not really, but it is the usual culprit `test-console.js`. It always behave quite randomly.  
Probably it should be investigated on its own.

Issue #535, opened 2024-06-04
Macros not compatible with C++


KaruroChori:
I am not too good with C, so I might be missing something; is there any good reason why these macros

```
#define TJS_CFUNC_DEF(name, length, func1)                                                                             \
    {                                                                                                                  \
        name, JS_PROP_C_W_E, JS_DEF_CFUNC, 0, .u = {.func = { length, JS_CFUNC_generic, { .generic = func1 } } }       \
    }
#define TJS_CFUNC_MAGIC_DEF(name, length, func1, magic)                                                                \
    {                                                                                                                  \
        name, JS_PROP_C_W_E, JS_DEF_CFUNC, magic, .u = {                                                               \
            .func = { length, JS_CFUNC_generic_magic, { .generic_magic = func1 } }                                     \
        }                                                                                                              \
    }
#define TJS_CGETSET_DEF(name, fgetter, fsetter)                                                                        \
    {                                                                                                                  \
        name, JS_PROP_C_W_E, JS_DEF_CGETSET, 0, .u = {                                                                 \
            .getset = { .get = { .getter = fgetter }, .set = { .setter = fsetter } }                                   \
        }                                                                                                              \
    }

```

cannot be written as

```
#define TJS_CFUNC_DEF(name, length, func1)                                                                             \
    {                                                                                                                  \
        name, JS_PROP_C_W_E, JS_DEF_CFUNC, 0,  {.func = { length, JS_CFUNC_generic, { .generic = func1 } } }           \
    }
#define TJS_CFUNC_MAGIC_DEF(name, length, func1, magic)                                                                \
    {                                                                                                                  \
        name, JS_PROP_C_W_E, JS_DEF_CFUNC, magic, {                                                                    \
            .func = { length, JS_CFUNC_generic_magic, { .generic_magic = func1 } }                                     \
        }                                                                                                              \
    }
#define TJS_CGETSET_DEF(name, fgetter, fsetter)                                                                        \
    {                                                                                                                  \
        name, JS_PROP_C_W_E, JS_DEF_CGETSET, 0, {                                                                      \
            .getset = { .get = { .getter = fgetter }, .set = { .setter = fsetter } }                                   \
        }                                                                                                              \
    }

```
? The underlying struct has those fields already specified in such order, so I cannot see that `.u` being helpful for C.
C++ cannot accept that syntax because either all labels are specified or none.  
So the second version should be portable on both languages in theory.

saghul:
[...]

PR #546, opened 2024-06-13
Support for C++ in macros

KaruroChori:
Fixing https://github.com/saghul/txiki.js/issues/535


KaruroChori:
Strangely `make format` does not work on my system and returns a `No such file or directory`, but running it manually does.
I guess my env variables are messed up.

Issue #479, opened 2024-04-09
URL.createObjectURL


EmixamPP:
Hey, I would like to ask your help concerning the missing `URL.createObjectURL` browser feature.

Here's some background, I would like to use the MQTT.js library (https://github.com/mqttjs/MQTT.js), which is compatible with browsers. However, I encountered two problems:
1. The lib detects whether it's in the browser by checking whether a `document` variable is defined in the global scope (because it is also compatible with Node.js by using the ws lib instead of "native" WebSocket). Thus I just added:
   ```
   Object.defineProperty(globalThis, 'document', {
    enumerable: true,
    get() {
        return globalThis;
    },
    set() {}
   });
   ```
   in the `src/js/polyfills/global.js` file
2. The lib requires the `URL.createObjectURL` https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL_static and `URL.revokeObjectURL` https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL_static functions

This is where I ask for your help, do you think there is a way to make this feature work in `txiki.js`?

Here is what I've tried (this is really naïve); I've added:
```
const chars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz';
function generateString(length) {
    let result = '';
    for (let i = 0; i < length; ++i) {
        result += chars.charAt(Math.floor(Math.random() * chars.length));
    }
    return result;
}

URL.createObjectURL = (obj) => {
    const url = generateString(32);
    globalThis[url] = obj;
    return url;
};

URL.revokeObjectURL = (url) => {
    delete globalThis[url];
};
```
to `src/js/polyfills/url.js`

After that, the lib does not require additional feature, and even work in some way, but it logs the following error:
`ReferenceError: could not load '<the random string generated>' - ENOENT: no such file or directory`

saghul:
I suspect the problem is that the URL represents a FIle or Blob, and it tries to read the file, which fails with a nu such file error.

Do you know where in mqtt.js the failure occurs?

EmixamPP:
Correct, this is a Blob

I tried to find from where this code comes from, and it seems it is imported from the worker-timer lib: https://github.com/chrisguttandin/worker-timers The object url is created by this function:
[...]

PR #548, opened 2024-06-17
ObjectURL api

EmixamPP:
ObjectURL API doc: https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL_static

It is more complicated than expected to solve #479.

I wanted to use the JS [`Blob.text()`](https://github.com/saghul/txiki.js/blob/69fa6a055cca4d6d8315fdeef35dcdd6bbd90d97/src/js/polyfills/blob.js#L118) method to retrieve the content of the stored blob (as a `JSValue`) in a global blob store. But the function is `async`, and I'm not sure the [`JS_Call`](https://github.com/saghul/txiki.js/pull/548/files#diff-b7d97be9c98fd50d49b9f2d6ee9bd3e4da589e223eeb0a06824d7af57d23efe6R516) method can resolve and wait such a function, since it raises `tjs__promise_rejection_tracker: Assertion '(JS_IsException(event)) == (0)' failed.`. 

Anyway, here is a draft PR that shows what I've tried so far, maybe you will be able to help me. Or my solution is too bad (I have the feeling that it is maybe a bit too hacky?), and it will be easier for you to start from scratch when you will have time later (don't worry if this is the case).

While I'm waiting for your review, I'll keep trying to fix and improve the solution.


saghul:
Good start!

What I had in mind was perhaps a bit simpler:

- Use the builtin SQLite module to store the Blob (probably in parts?) in an in-memory table
- Use `crypto.randomUUID()` for the identifier

Then when creating the worker you can grab all the Blob parts straight from the SQLite DB and re-construct the content. SInce the sqlite module is sync there is no need to await anyhing.

EmixamPP:
Sorry, I had to push force because I authored then committed with a wrong email.

So I got the api almost working in every case. Actually, in the test, I've added the revoke of the url after creating the worker. However, if you execute it a lot, sometimes, it will be done a bit too soon; before being queried by the eval. I will fix that.

I've used the `crypto.randomUUID()` for the url generation directly at the js side. And I will investigate then to use the sql module instead of using a global hash map as blob store. 

I've made the new commits as fixup for their previous related commit, like this when the review is done, I can squash them automatically to keep a clean commit history, while being able to review them easily.

saghul:
Feel free to keep pushing changes, we can squash when merging.

tests/test-worker-url.js
```diff
@@ -0,0 +1,18 @@
+import assert from 'tjs:assert';
+
+const data = JSON.stringify({ foo: 42, bar: 'baz!' });
+const worker = `if (self.addSignalListener || self.removeSignalListener) {throw new Error('There are signals in a Worker!');} self.postMessage(${data});`
+const blob = new Blob([worker], { type: 'text/javascript' });
+const url = URL.createObjectURL(blob);
+
+const w = new Worker(url);
+URL.revokeObjectURL(url);
```
EmixamPP, line 9: There is a problem that I cannot solve through the api or any tricks. It can happen that the `URL.revokeObjectURL`               is processed before the creation of the worker. Consequently, the blob will not be found when it is required during the `TJS_EvalModule`.

I'm a bit stuck at that point.

EmixamPP: Indeed, I could instead put the work in the Worker instead of the EvalModule. That was to keep intact the Worker polyfill.

saghul: I reckon it's a safer option!

EmixamPP:
I have the feeling that going for an in-memory sqlite database may be overkill, so here I simply use a js `Map()`. Also, mainly because, after a re-read of the api here: https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL_static. I read that the liftime of the url is linked to the scope of the document (in the browser), so globally scoped across runtime object is unnecessary. Let me know what you think about that.

I got a working implementation. Now, maybe some work on the style and structure are needed?
E.g. I made the use of a magic string between `url.js` and `objecturl.c` to know the name of the storage map. So, it may be better to create the storage map from C during the loading of the core module, but then, even if the API is not used, an empty map will be created. 

I wanted to use the `sessionStorage` but that one only store string, so that is a problem here, since the string value cannot be retrieved synchronously in js.

src/js/polyfills/url.js
```diff
@@ -4,3 +4,31 @@ import { URL, URLSearchParams } from 'whatwg-url';
...
+        }
+        return _objectURLs;
+    }
+});
+
+// TODO: straggly, without the wrapped console.log, the runtime will abort
+// more investigation is needed
+globalThis.URL.createObjectURL = (object) => {
+    console.log('');
+    if  (String(object) !== '[object Blob]') {
```
saghul, line 24: I think it's fine to do an isinstance check here.

src/js/polyfills/url.js
```diff
@@ -4,3 +4,31 @@ import { URL, URLSearchParams } from 'whatwg-url';
...
+        throw new TypeError("URL.createObjectURL: Argument 1 is not valid for any of the 1-argument overloads.");
+    }
+    const url = `blob:${crypto.randomUUID()}`;
+    objectURLs.set(url, object.text()); // store the promise, will computed at most once
```
saghul, line 28: What if we want the binary data?

src/worker.c
```diff
@@ -119,7 +127,14 @@ static void worker_entry(void *arg) {
 
     /* Load the file and eval the file when the loop runs. */
     JSValue filename = JS_NewString(ctx, wd->path);
-    CHECK_EQ(JS_EnqueueJob(ctx, worker_eval, 1, (JSValue *) &filename), 0);
+    JSValue blob = JS_UNDEFINED;
+    if (wd->blob)
```
saghul, line 131: Where is this set?

saghul (changes requested):

saghul:
Thank you! I did a full review round and left some comments.

src/vm.c
```diff
@@ -483,6 +483,26 @@ int tjs__load_file(JSContext *ctx, DynBuf *dbuf, const char *filename) {
     return r;
 }
 
+JSValue TJS_EvalModuleContent(JSContext *ctx, const char *filename, bool is_main, bool use_real_path, const char *content, size_t len) {
```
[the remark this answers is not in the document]
EmixamPP: I just moved here to some code from `TJS_EvalModule` to a separate function, because the same code is needed to evaluate the blob inside the ObjectURL as a module. Because the only difference is that we do need to load the content from a file.

src/vm.c
```diff
@@ -483,6 +483,26 @@ int tjs__load_file(JSContext *ctx, DynBuf *dbuf, const char *filename) {
...
+    /* Compile then run to be able to set import.meta */
+    JSValue ret = JS_Eval(ctx, content, len, filename, JS_EVAL_TYPE_MODULE | JS_EVAL_FLAG_COMPILE_ONLY);
+    if (!JS_IsException(ret)) {
+        js_module_set_import_meta(ctx, ret, use_real_path, is_main);
+        ret = JS_EvalFunction(ctx, ret);
+    }
+
+    /* Emit window 'load' event. */
```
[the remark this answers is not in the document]
EmixamPP: Just moved some code from below, since it is executed when loading a worker from a file, so I guess that should also be the case if the content is in a variable?

src/worker.c
EmixamPP: In the constructor of the worker, when creating the `worker_data`

src/js/polyfills/url.js
```diff
@@ -4,3 +4,31 @@ import { URL, URLSearchParams } from 'whatwg-url';
...
+    configurable: false,
+    get: () => {
+        if (!_objectURLs) {
+            _objectURLs = new Map();
+        }
+        return _objectURLs;
+    }
+});
+
+// TODO: straggly, without the wrapped console.log, the runtime will abort
```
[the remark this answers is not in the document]
EmixamPP: The problem is different now, the function is spammed. I made a small video capture to show you:
[objecturl_bug.webm](https://github.com/saghul/txiki.js/assets/37633694/c1911300-1bd9-4657-9072-0870b301cc1b)
Therefore, with the current case, I have no backtrace to provide. And I have no clue what is that printed SymtaxError, since there is no problem at all when running it with the empty `console.log('')`.

src/js/polyfills/worker.js
```diff
@@ -9,29 +9,41 @@ class Worker extends EventTarget {
     constructor(path) {
         super();
 
-        const worker = new _Worker(path);
+        const resolve_worker = (resolve, path, blob) => {
```
saghul, line 12: After seeing how this turned out I'm afraid we'd deviate from the standard semantics, for example with regards to terminate.

So, here is another approach, which is somewhat what Deno does:

Use a hidden Symbol to get the parts out of a Blob, as an array. Then you can work with those and create the code string out of them syncrhonously!

saghul: In addition, not in this PR, we should support data URLs, so what today is called `path` should really be `specifier`, as the standard says.

saghul (changes requested):

src/js/polyfills/worker.js
EmixamPP: Is what I've done correct? I'm not sure what I'm supposed to do when it comes to hidden symbols. And I thought we could directly create a method for converting text synchronously, instead of accessing the parts and then converting them; to put the conversion work inside the Blob class and in each class that requires it.

saghul: Not really. You either define the symbol somewhere and import it in both places or you define a global named symbol like so: `Symbol.for('tjs.internal.blob.get-parts')` for example. So if you use it like that in both places, you'll get the same symbol.

EmixamPP: Ok I will do the changes. However, is the fact that this is a function returning the text ok? Or you prefer the parts? 
PS: Note that what I did works perfectly (I don't know if this is supposed to be the case).

saghul: Also, when converting to text use TextDecoder, there might be utf-8 inside.

EmixamPP: I hope, this time, I'm right!

src/js/polyfills/url.js
```diff
@@ -4,3 +4,24 @@ import { URL, URLSearchParams } from 'whatwg-url';
...
+
+    const url = `blob:${crypto.randomUUID()}`;
+
+    _objectURLs.set(url, object);
+
+    return url;
+};
+
+globalThis.URL.revokeObjectURL = url => _objectURLs.delete(url);
+globalThis.URL.getObjectURL = url => _objectURLs.get(url);
```
saghul, line 27: THis also needs to be hidden with some symbol like the blob. thing.

src/js/polyfills/worker.js
```diff
@@ -2,14 +2,34 @@ const core = globalThis[Symbol.for('tjs.internal.core')];
...
+
+    for (const part of blob[kBlobGetParts]) {
+        if (part instanceof Blob) {
+            str += blobTextSync(part);
+        } else {
+            str += decoder.decode(part, { stream: true });
+        }
+    }
+
+    str += decoder.decode();
```
saghul, line 21: IIRC it's better to collect all parts in an array and then join it with ''

EmixamPP: I've just copied the implementation used in `Blob.text`

src/js/polyfills/blob.js
```diff
@@ -119,16 +120,16 @@ class Blob {
         // More optimized than using this.arrayBuffer()
         // that requires twice as much ram
         const decoder = new TextDecoder();
-        let str = '';
+        let partsStr = [];
```
saghul, line 123: you can make this const now.

src/worker.c
```diff
@@ -44,6 +44,7 @@ static JSClassID tjs_worker_class_id;
 
 typedef struct {
     const char *path;
+    const char *blob;
```
saghul, line 47: Let's call this source, since we will extend it to support data URLs too.

src/worker.c
```diff
@@ -77,7 +78,13 @@ static JSValue worker_eval(JSContext *ctx, int argc, JSValue *argv) {
         goto error;
     }
 
-    ret = TJS_EvalModule(ctx, filename, false);
+    if (!JS_IsUndefined(argv[1])) {
+        const char *blob = JS_ToCString(ctx, argv[1]);
```
saghul, line 82: Use `JS_ToCStringLen` so you get the length straight and can avoid doing `strlen`.

src/worker.c
```diff
@@ -119,7 +126,12 @@ static void worker_entry(void *arg) {
 
     /* Load the file and eval the file when the loop runs. */
     JSValue filename = JS_NewString(ctx, wd->path);
-    CHECK_EQ(JS_EnqueueJob(ctx, worker_eval, 1, (JSValue *) &filename), 0);
+    JSValue blob = wd->blob ? JS_NewString(ctx, wd->blob) : JS_UNDEFINED;
```
saghul, line 129: Ditto, let's call this source.

src/js/polyfills/blob.js
```diff
@@ -242,3 +247,5 @@ Object.defineProperty(window, 'Blob', {
     writable: true,
     value: Blob
 });
+
+export { kBlobGetParts };
```
saghul, line 251: Since this file changes the global already, importing it at the wrong time might have implications. Thus I think it's best to use a global symbol. `Symbol.for('tjs.internal.blob.getParts')`.

If you use that on both places it will return the same symbol instance, so no imports are necessary.

src/js/polyfills/worker.js
```diff
@@ -1,15 +1,42 @@
...
 class Worker extends EventTarget {
     constructor(path) {
         super();
 
-        const worker = new _Worker(path);
+        let source = undefined;
+
+        const blob = URL[kGetObjectURL](path);
+
+        if (blob) {
```
EmixamPP, line 35: I am wondering if for the readability, and preparing your future support for data URL, it wouldn't be better to have a function to check if path is an objecturl, if so, then get the objecturl. In the end, this adds an unnecessary step, but it may make the code more expressive?

saghul: I agree. Basically parse the path as a URL, and check the scheme.

src/js/polyfills/blob.js
saghul: You don't need to put them anywhere. Wherever you need it, just use `Symbol.for('tjs.internal.blob.getParts')` and you'll get back the same reference.

src/js/polyfills/worker.js
EmixamPP: Hum, a problem is that currently, when passing a path to the worker, since it does not have an url pattern, putting that path inside a new URL object raise an error. So, as an alternative, I used `URLPattern`.

saghul: That is fine! You'll want to do something like:

```
let url;
try {
    url = new URL(specifier);
} catch (_) {
    // It's not a URL, it's a path.
}
...
```

saghul:
@EmixamPP I discovered the source of the problem! The `fflate` package which we use as part of the compression streams polyfill was using the Worker API while the engine was not fully ready, in order to detect support.

I disabled it since it didn't actually work, and squashed and force-pushed!

saghul (approved):

Issue #541, opened 2024-06-08
 `test-console.js` randomly failing


KaruroChori:
> Not really, but it is the usual culprit `test-console.js`. It always behave quite randomly.  
Probably it should be investigated on its own.

_Originally posted by @KaruroChori in https://github.com/saghul/txiki.js/issues/537#issuecomment-2156029100_

PR #572, opened 2024-07-01
test: make test-console more resilient

saghul:
Read from stdout/stderr while waiting for the process to end.

Fixes: https://github.com/saghul/txiki.js/issues/541


saghul:
FML, I cannot repro the bug locally but the CI seems to be able to do it at least 25% of the time :-/

saghul:
Alright, I ran the test 1000 times in a row locally and I think we're good now!

Issue #160, opened 2021-02-08
qjsc-like compiler


reaktivo:
Would it be possible to have a qjsc-like compiler built into txiki that would allow bundling user code with the txiki runtime?

saghul:
Yeah. Ideally txikijs would do it itself, with some -c flag for example...

ArtDmn:
Is it still something investigated ?

saghul:
It is something I'd like to eventually do yeah.

Tobbe:
If I start writing something for NodeJS or Bun today, is there something I should watch out for, that would make it difficult/impossible to then later move it to txiki.js? When/if this eventually gets implemented here, of course.

saghul:
Not sure I understand the question :-)

If you write a script without any runtime-specific API it should, in principle, work across all of them.

Tobbe:
> If you write a script without any runtime-specific API it should, in principle, work across all of them.

When you wrote this I didn't really understand what implications this had. Now, with the discussions in https://github.com/saghul/txiki.js/discussions/525 and https://github.com/saghul/txiki.js/discussions/515, I realize just how big of a hurdle this would be to get things I'm used to writing working with txiki.js 😅   But that definitely does not mean I'm deterred from trying to use txiki.js for what I initially wanted to do with it. If anything I'm even more excited to try to see what I can build as soon as you have some time to spend on `tjs compile` functionality 😍

MulverineX:
Ideally, this should also have an option to compile to non-executable bytecode _without_ bundling the runtime.

saghul:
Indeed.

saghul:
I was looking into this just today, and I think there are 2 ways it can go:

Option A:

`tjs compile file.js`

That would bytecompile `file.js` and use a similat trick to Deno: create a new executable which is a copy of the current one, then add the bytecode, an index and a trailer. When the self-contained executable starts, it will check if it ends in the trailer and if so, get the index, and evaluate the bundled bytecode.

The cons of this is that there needs to be a bundling process, in order to bytecompile the self-contained executable without external dependencies in imports and the like. Dynamic imports might be challening.

Option B:

`tjs compile directory`

The directory must contain an `index.js` file.

Here we'd (ab)use the ZIP file format. The trick is to compress the directory, and simply append the resulting ZIP file to a copy of the current executable.

Then when the executable is launched, it will check if it embeds a ZIP file, if it does, extract and run `index.js` .

[...]

PR #573, opened 2024-07-01
core: introduce the ability to create standalone binaries

saghul:
The compile, serialize, deserialize and evalBytecode primitives as exposed (for now) so other scenarios such as writing bytecode to a file and evaluating it can be explored.

They are not documented on purpose ;-)

Closes: https://github.com/saghul/txiki.js/issues/160


Tobbe:
Wow! This is amazing! Thanks for making this happen ❤️ 

Can't wait to start playing around with compiling binaries!

Tobbe:
image

🎉

PR #583, opened 2024-07-02
Clean tjs namespace, round 1

saghul:


saghul:
@KaruroChori Please take a look! Thank you for making me think about this again. There was certaily a bunch of stuff that was easily actionable.

Now on to the next steps. I do think _some_ namespacing can be done!

- vm: serialize / deserialize / compile / evalBytecode / gc / versions? ( this one is pretty clear I think!)
- hwInfo: availableParallelism, cpus, networkInterfaces

There are some items that seem to belong somewhat together, but I can't quite put my finger on it: platform / what's in uname / uptime / loadavg / memory-info?

Last, some other random thoughts:

- PosixSocket is a low level API not to be used by most, so it probably belongs in some `tjs:posix-socket` module, that would move many constants too
- Abstract getaddrinfo into some lookup() function a la Node

KaruroChori:
That's great, I think this greatly improves the library. I will need a bit to check its details but overall it looks good.

> There are some items that seem to belong somewhat together, but I can't quite put my finger on it: platform / what's in uname / uptime / loadavg / memory-info?

All this was stuffed in `tjs.platform` in my proposal, alongside the content of what you suggested to have in `hwInfo`. I consider them to be all platform-specific details so it did make sense for me to have them all together.

Your `vm` was basically my proposed `engine`, so I am very much supportive of this as well, and I like the presence of even more features in it compared to what I originally considered.

> PosixSocket is a low level API not to be used by most, so it probably belongs in some tjs:posix-socket module, that would move many constants too
> Abstract getaddrinfo into some lookup() function a la Node

I agree too

docs/types/txikijs.d.ts
```diff
@@ -944,17 +929,17 @@ declare global {
         const ppid: number;
         
         interface UserInfo {
-            username: string;
+            userName: string;
             uid: number;
```
KaruroChori, line 933: For consistency maybe having `userId` and `groupId` in place of those shorts acronyms would be better

docs/types/txikijs.d.ts
```diff
@@ -374,14 +374,7 @@ declare global {
...
         * @param path Path to convert.
         */
-        function realpath(path: string): Promise<string>;
-        
-        /**
-        * Removes the given file.
-        *
-        * @param path Path to be removed.
-        */
-        function unlink(path: string): Promise<void>;
```
KaruroChori, line 384: There is a difference between `remove` and `unlink` which most people would not be able to state without checking the docs :D. So it does make sense to remove one & simplify the interface since in the vast majority of cases one is redundant, but I just wanted to make sure this was an "informed choice".

docs/types/txikijs.d.ts
```diff
@@ -599,24 +585,16 @@ declare global {
         * @param path The path to of the directory to be created.
         * @param options Options for making the directory.
         */
-        function mkdir(path: string, options?: MkdirOptions): Promise<void>;
+        function makeDir(path: string, options?: MkdirOptions): Promise<void>;
         
         /**
         * Copies the source file into the target.
         *
-        * If `COPYFILE_EXCL` is specified the operation will fail if the target exists.
```
KaruroChori, line 607: I am not supportive of removing these flags as they have utility. I would rather add configuration flags in `CopyFileOptions` to reflect them.

docs/types/txikijs.d.ts
```diff
@@ -579,17 +572,10 @@ declare global {
...
         function open(path: string, flags: string, mode?: number): Promise<FileHandle>;
         
-        /**
-        * Removes the directory at the given path.
-        *
-        * @param path Directory path.
-        */
-        function rmdir(path: string): Promise<void>;
-        
         interface MkdirOptions {
```
KaruroChori, line 575: Probably it should be named MakeDirOptions

saghul:
> That's great, I think this greatly improves the library. I will need a bit to check its details but overall it looks good.

Sweet. I'll likely merge this and continue with the next steps and refine as we go.

> All this was stuffed in `tjs.platform` in my proposal, alongside the content of what you suggested to have in `hwInfo`. I consider them to be all platform-specific details so it did make sense for me to have them all together.

I think thay makes sense. The current `tjs.platform` is pretty much the `sysname` field of `uname(2)` so it can be extracted from there for sure.

src/os.c
```diff
@@ -492,18 +492,18 @@ static const JSCFunctionListEntry tjs_os_funcs[] = {
...
-    TJS_CFUNC_DEF("tmpdir", 0, tjs_tmpdir),
     TJS_CFUNC_DEF("random", 3, tjs_random),
     TJS_CFUNC_DEF("cpuInfo", 0, tjs_cpu_info),
     TJS_CFUNC_DEF("loadavg", 0, tjs_loadavg),
     TJS_CFUNC_DEF("networkInterfaces", 0, tjs_network_interfaces),
-    TJS_CFUNC_DEF("gethostname", 0, tjs_gethostname),
-    TJS_CFUNC_DEF("getPid", 0, tjs_getpid),
-    TJS_CFUNC_DEF("getPpid", 0, tjs_getppid),
-    TJS_CFUNC_DEF("userInfo", 0, tjs_userInfo),
     TJS_CFUNC_DEF("availableParallelism", 0, tjs_availableParallelism),
```
KaruroChori, line 499: Personally I would have called it `threads` because it is a bit of a mouthful, but that is just a very subjective opinion :D.

src/js/core/index.js
```diff
@@ -33,26 +33,31 @@ const noExport = [
     'evalFile',
     'evalScript',
     'ffi_load_native',
+    'getPid',
+    'getPpid',
     'guessHandle',
     'isStdinTty',
+    'isWorker',
```
KaruroChori, line 40: Making this not visible, is there an alternative web compliant mechanism working to detect if it is a worker?

KaruroChori (commented):
Full review

docs/types/txikijs.d.ts
saghul: Yes this is intentional. It's an extension of remove, which supports recursive deletion, off by default.

docs/types/txikijs.d.ts
saghul: 👍

docs/types/txikijs.d.ts
saghul: I have never needed them and they are so low level I'm not sure they have a use in the context where txiki.js is used. Adding an options object afterwards will be no problem.

docs/types/txikijs.d.ts
saghul: 👍

src/js/core/index.js
saghul: Yes, self.postMessage for example. But also, I should stop defining self on the main, self only has meaning in a worker scope.

src/os.c
KaruroChori: I was referring to hardware threads, not software ones.
Like Sockets > Cores > Threads > Pipeline.  
Usually the parallelism of a system is linked to threads since they are the latests stages capable of virtually performing all operations of the instruction set (with many asterisks).  
But yes, if this nomenclature creates ambiguity maybe it is better to leave it as it is :D.

PR #592, opened 2024-07-08
core: introduce tjs.engine

saghul:
It includes the bytecode manipulation functions in addition to the versions object.

This will be the namespace to add engine specific stuff, like the engine memory usage, for example.

~~TODO: docs.~~


KaruroChori:
This regressed/reverted the gc object PR I added few weeks ago features, but tests have not been changed to show that.
Either way it should be fixed.

KaruroChori:
Also, the documentation generated from ts.d will only nest children on the left tree view for `namespaces`. It might be worth to declare `tjs.engine` as such, so that its first-level entries will be shown directly on the menu.

saghul:
> The interface in the types file is incomplete for some reason which threw me off. Might it be that I missed to push a commit on that occasion?

Probably yeah!

> Also, the documentation generated from `ts.d` files will only nest children on the left tree view for `namespaces`, not for interfaces. It might be worth to declare `tjs.engine` as such, so that its first-level entries will be shown directly on the menu.

Hum, I'll take a look at that, thanks for the heads up!

saghul:
Screenshot 2024-07-08 at 16 44 46

Looks like it's OK? I see the interfaces as members of the tjs namespace on the left bar.

KaruroChori:
Yes, what I was saying is that its children are not shown as a nested menu.
Since `tjs.engine` is effectively just serving as namespace, it would have made sense to handle it as such from a typing perspective, and so tsdoc would have generated a subfolder in the menu for its children.

saghul:
Gotcha. Since that was kinda there already, I'll take a look at it all together in a separate PR 👍

PR #624, opened 2024-07-20
Document env var & make generalized history db

KaruroChori:
I documented the environment variable. Probably we should plan a better way to handle documentation of the project for everything which is not trivially `.d.ts` related.
I also generalized the naming around history db, as I plan to add more tables to support additional functionality in it.


saghul:
Let's separate the changes please.

Also, renaming the DB before we are adding new functionality doesn't add much so I'd rather do it in a single shot.

Session as a name is also too generic and perhaps confusing since some might think it's related to session storage somehow.

How about repl.db since that's really what it will have? Stuff that you work on on the REPL, right?

saghul:
Regarding the docs, I agree!

I have been tinkering with a type doc plugin for Docusaurus and I hope to go for it very soon!

KaruroChori:
My plan was to add in there more than just stuff for the repl:
- Bytecode cache for modules already compiled once in the past.
- Secrets.
History in this case would only be relevant for the repl, but the other two would be shared. So `repl.db` would not be good for what I am trying to implement.

> I have been tinkering with a type doc plugin for Docusaurus and I hope to go for it very soon!

:+1:

saghul:
Interesting idea! I don't want to mix that with the history database though. We could certainly try it out on a different DB.

KaruroChori:
For reference, I wrote in https://github.com/saghul/txiki.js/discussions/628 my ideas for the cache thing.

Issue #566, opened 2024-06-29
sqlite extensions


KaruroChori:
Sqlite has a native mechanism to handle extensions to be added at runtime. They can either be added via [`sqlite3_load_extension`](https://www.sqlite.org/c3ref/load_extension.html) or with the sql function [`load_extension(X,Y)`](https://www.sqlite.org/lang_corefunc.html#load_extension).
In the first case we need to enable this functionality configuring the db handle as shown in the referenced page, in the second case by using [`sqlite3_enable_load_extension`](https://www.sqlite.org/c3ref/enable_load_extension.html) which also enables the sql function.

I think it would be nice to have support for sqlite native extensions in txiki.

saghul:
How does that work with regard to linking to the right sqlite library?

KaruroChori:
It is my understanding that, while extensions can be statically linked if so desired, the general approach is for them to be dynamically linked at runtime by the sqlite3_load_extension.
This will be slighly different for each operating system, but on linux it is likely a wrapper to dlopen and a callback to the init function each extension must provide.

Distributing the extensions and making sure they are compatible with the sqlite in txiki is a concern for who develops the final application, not of txiki.
As we already provide the sqlite version in `tjs.versions()` and any relevant platform information, there is nothing more that the runtime is expected to do in my opinion (other than exposing those functions to load them).

saghul:
What I mean is that txiki statically links sqlite with certain build flags.

What version are the ones building the extension going to link against?

Is there a chance for incompatibilities or is ABI stable enough?

Sorry for the many questions, it's the first time I tackle this 😅

KaruroChori:
For most scenarios, I would expect extensions compiled for the same architecture to work, regardless of the flags set by any specific sqlite distribution.
Obviously, this will not be the case if specific features have been disabled and they are explicitly needed by any extension. But I do not expect it to be a problem related to incompatible ABIs for most configurations. I might be wrong on this.

That being said, as I expressed before, this is not a concern of txiki, but of whoever is shipping the final application. As long as the flags to build sqlite are reported in txiki's sourcecode (and possibly documented), the distribution & validation of those extensions is beyond our control.

KaruroChori:
Basically: https://github.com/cloudmeter/sqlite/blob/master/sqlite3ext.h
This is stable enough.

saghul:
Cool, I can give this a try!

saghul:
Fixed in https://github.com/saghul/txiki.js/pull/627

PR #627, opened 2024-07-20
Added loading of sql extensions

KaruroChori:
Closing https://github.com/saghul/txiki.js/issues/566
I tested it on my system, and it loads the extension properly.  
Adding a proper testsuite would require us to generate a minimal demo extension as part of the build process for each arch, to load it and run few demo queries against.

It is my understanding a similar approach has been followed for the ffi module.


KaruroChori:
Minimal testing added. Hopefully it will work on all supported architectures.

saghul:
Thanks for walking the extra mile, this is awesome! I'll get it reviewed shortly.

Issue #630, opened 2024-07-21
Encoding not supported. Only utf-8 is supported


ahaoboy:
There are many libraries that use ``` new TextDecoder("utf8")```

https://github.com/arv/text-encoding-utf-8/issues/4

An alternative library is text-decoding, but it is 500k
https://github.com/idiocc/text-decoding

saghul:
Our TextEncoder/Decoder implementation only supports UTF8.

I have no plans to change this since it adds a significant chunk to our bundle size.

Do you get a problem when instantiating it with that name?

ahaoboy:
> Do you get a problem when instantiating it with that name?

Both of these are valid utf8 formats, some libraries mix the two and cause errors

```
new TextDecoder("utf8")

new TextDecoder("utf-8")
```

text-encoding-utf-8 is no longer maintained, so it can only be fixed by patch or polyfill. This is my current temporary solution
```
import { TextEncoder, TextDecoder as BaseTextDecoder } from 'text-encoding-utf-8';

window.TextEncoder = TextEncoder;
window.TextDecoder = class TextDecoder extends BaseTextDecoder {
  constructor(encoding, options) {
    encoding = String(encoding).toLowerCase()
    encoding = encoding === 'utf8' ? 'utf-8' : encoding
    super(encoding, options)
  }
};
```

KaruroChori:
Looking at the specs reported in the issue you linked, while there might be some slight differences amongst those entries, for maximum compatibility we should just provide any of the following as alias names for utf8:
```
"unicode-1-1-utf-8"
"unicode11utf8"
"unicode20utf8"
"utf-8"
"utf8"
"x-unicode20utf8" 
```
It would require no further effort on the encoding/decoding side.

ahaoboy:
> It would require no further effort on the encoding/decoding side.


There are indeed many aliases, but I have only encountered utf8 and utf-8 
It may also be due to the optimization of the bundle tool? utf8 will have one less character than utf-8

KaruroChori:
I think this has more to do with the fact we had breaking changes to the encoding between unicode 1.1 and 2. 
If memory serves me right, the hangul from korean were changed places (breaking) and tibetan was first removed and later reintroduced.

So documents encoded in those years for these respective languages are a mess to handle, unless they have been later re-encoded.
[...]

PR #632, opened 2024-07-23
add patch for text-encoding-utf-8, support utf8

ahaoboy:
close: https://github.com/saghul/txiki.js/issues/630
If there is a better solution, please close this PR


ahaoboy:
Why use window instead of global This as the global object? Is it because events can be listened on window?

saghul:
We already use patch-package on the polyfill, can you please apply the patch there instead?

saghul:
> Why use window instead of globalThis as the global object? Is it because events can be listened on window?

Out of habit, but yeah this should be globalThis.

ahaoboy:
> We already use patch-package on the polyfill, can you please apply the patch there instead?

done

ahaoboy:
The behavior of patch-package is a bit strange, I may need to investigate further, and it seems that after installing patch, the compilation result of make js && make does not change, which is a bit strange

saghul:
Yeah when the polyfills change the Makefile doesn't quite pick it up because the dependencies across files are not detected.

You can force it by deleting the .c file.

ahaoboy:
What's strange is that this patch doesn't seem to modify the source code, so why is it added here?
Is it because the editor automatically formats it?
```patch
@@ -639,4 +640,4 @@ function UTF8Encoder(options) {
 }
 
 exports.TextEncoder = TextEncoder;
-exports.TextDecoder = TextDecoder;
\ No newline at end of file
+exports.TextDecoder = TextDecoder;
```

I only added an alias utf8, which should cover most usage scenarios.

saghul:
Thank you!

PR #647, opened 2024-08-21
polyfills: bundle our own fetch polyfill

saghul:
It's vendored and slightly adapted whatwg-fetch to start with.


saghul:
Alright, here is the plan. This is not fully  spec compliant yet, but it is better tailored for our needs since it's easier to modify.

I think the next step would be to create a generic HTTP client which is not XHR or fetch and use it from both directly. This would detach the fetch implementation from XHR and allow us to add the missing stream features. In addition it should simplify the XHR implementation by moving big parts of it to JS. Last, if / when the HTTP library is replaced only one thing needs to be replaced.

PR #652, opened 2024-09-21
Prep work for WAMR integration

KaruroChori:
**Not to be merged, just to track what I am doing.**

In short, wasm3 is no longer maintained and very slow due to its interpreted nature.
I think it is best for txiki to move away from it and I would suggest to try https://github.com/bytecodealliance/wasm-micro-runtime.

It is advertised as highly configurable and embeddable. https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/doc/embed_wamr.md
One can selectively decide which feature are wanted, limiting the final footprint. It is also possible to run wasm without allocations outside a preallocated region, making it suitable for real embedded applications.
Beyond that, it has a much wider support of standards compared to wasm3, and a AOT can be enabled for systems if they can afford its overhead.

Moving to WAMR would enable further developments in areas where txiki is currently stuck, like adding a fast transpiler to semi-natively run typescript and jsx code.


KaruroChori:
By the way, I would suggest to disable the automatic pipelines from running for draft PR.

saghul:
Thanks for getting this effort started! ❤️‍🔥

saghul:
> By the way, I would suggest to disable the automatic pipelines from running for draft PR.

Not sure how to do that, but I'd be happy to turn that on!

KaruroChori:
https://github.com/orgs/community/discussions/25722
There is a flag to be added to the workflow to either check the draft status or activate it when set to "ready for review".

KaruroChori:
Just a quick update. The wasi version of esbuild from https://www.npmjs.com/package/@esbuild/wasi-preview1?activeTab=readme seems to be working fine on WAMR. wasip1 cannot open sockets, so some of the more obscure features of esbuild might not be available, but for the intended purpose it should be covered.

saghul:
That's awesome! Let's do it then!

KaruroChori:
Sort of, it is on the list of things to do.
I am job hunting at the moment and I had other projects taking priority, but I hope I find enough time to revisit this soon(TM).

Issue #653, opened 2024-09-21
ffi string not working as expected


KaruroChori:
```js
//import type from '@txikijs/types';

const sopath = './libcfltk.so';
import FFI from 'tjs:ffi';
const testlib = new FFI.Lib(sopath);

const cfltk ={
    Fl_Window_new: new FFI.CFunction(testlib.symbol('Fl_Window_new'), FFI.types.pointer, [FFI.types.sint,FFI.types.sint,FFI.types.sint,FFI.types.sint,FFI.types.string]),
    Fl_Window_end: new FFI.CFunction(testlib.symbol('Fl_Window_end'), FFI.types.void, [FFI.types.pointer]),
    Fl_Window_show: new FFI.CFunction(testlib.symbol('Fl_Window_show'), FFI.types.void, [FFI.types.pointer]),
    Fl_run: new FFI.CFunction(testlib.symbol('Fl_run'),FFI.types.void, [])
}

const handle = cfltk.Fl_Window_new.call(0, 0, 300, 400, "AAAAA");
cfltk.Fl_Window_end.call(handle);
cfltk.Fl_Window_show.call(handle);
cfltk.Fl_run.call();
```
The window should show "AAAAA" as title. The same (adapted) code run in Bun results in the expected behaviour. 
According to code I have seen in `test` this should be valid.

I attached the library in case you want to try it youself. It is derived from https://github.com/MoAlyousef/cfltk for linux amd64.
[libcfltk.so.gz](https://github.com/user-attachments/files/17085701/libcfltk.so.gz)

KaruroChori:
Surprisingly a testing it on a different library is fine:
```
const testlib2 = new FFI.Lib("./libsum.so");
const lib2={
    concatenateStrings: new FFI.CFunction(testlib2.symbol('concatenateStrings'), FFI.types.string, [FFI.types.string,FFI.types.string])
}
console.log(lib2.concatenateStrings.call("h","e"))
```
results in the expected "he".

KaruroChori:
It looks like a bug related to the lifetime of that string (the prototype is using pointer in place of string here).
This works
```js
const vuff=(new TextEncoder()).encode('hello\0');
const handle = cfltk.Fl_Window_new.call(0, 0, 300, 400,ffiInt.getArrayBufPtr(vuff) );
```

This does not
```js
[...]

PR #654, opened 2024-09-22
Solving issues in https://github.com/saghul/txiki.js/issues/653

KaruroChori:
This solves some issues I found out while debugging #653.
The current implementation of strings as an advanced type is extremely problematic.
What is really passed onto the C function is a clone of the original object with a shorter lifetime due to its scoping. This causes issues when that object is expected to be consumed later once the reference counter already collected it.

The new syntax is not as compact or nice, but there is no reasonable way to make the original better from what I can tell (without hacking the js syntax).
I already ported these changes to the test-suite and verified it worked.

There is one missing bit, at the moment `fromBuffer` for CString is very wasteful as it forces to decode and reencode.
There is surely a better way to bind the buffer directly, but I am not very familiar with the intrinsics provided in txiki.


saghul:
@lal12 mine taking a look?

lal12:
I don't think it is that big of an issue. It is not quite obvious what happens, but that if a C function expects the memory to be existent for a while you have to take special care, that is just an inherent thing with memory unsafe programming languages.

It also isn't quite "unreasonable" to expect such a behavior, since strings are primitives in JS and therefore effectively call-by-value.

This PR would be additional overhead and less comfortable for most use cases. 

There can an argument be made for the automatic behavior of the parser though, since it cannot really decide what type to use (advanced buffer, string or simple ptr). I never really intended it to be a full C prototype parser, just for an easier way for people to use ffi without having to learn an API, so you can control it by using `void*` instead of `char*`, e.g.. This of course isn't really applicable when using it with original header files.

lal12:
I guess allowing both a string or a buffer (or even a dedicated string class) to be passed, would be an option though.

KaruroChori:
> I don't think it is that big of an issue. It is not quite obvious what happens, but that if a C function expects the memory to be existent for a while you have to take special care, that is just an inherent thing with memory unsafe programming languages.

While I agree in principle, as this would apply to pointers and buffers alike, for string it is just made worse by the attempt of a cleaner syntax.
The current implementation is hiding `ctx.buf = (new TextEncoder()).encode(str+'\0');` on a narrower scope compared to the lifetime of the original string, and it will never be possible to keep it alive from the outside since its reference is immediately lost. Because of that, strings are intrinsically more broken compared to basic pointers, as the user would not even know why if they don't go and check the implementation in txiki:ffi.
If the lifetime is fine for everything but not for strings I would consider it a problem of the string implementation.

lal12:
Some information about this should obviously go into the documentation.

I am not fundamentally in disagreement, I guess it really depends on the use case, what in the end is more practical.

But I just noticed another issue in the current implementation and the PR. If you get a char* pointer as a return and need it later (e.g. for passing it to a free function), it also would be lost.

saghul:
What do other ffi APIs do here? For example Python's cffi? Perhaps we could borrow some inspiration from there.

PR #668, opened 2024-11-13
doc: add libltdl-dev for linux

ahaoboy:
Compiling on ubuntu24.04 with github runner will give an error

```
configure.ac:215: error: possibly undefined macro: LT_SYS_SYMBOL_USCORE
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
autoreconf: error: /usr/bin/autoconf failed with exit status: 1
```

https://github.com/libffi/libffi/issues/210

It seems that libuv and libffi require libltdl-dev

```
sudo apt-get install libltdl-dev
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,15 +127,15 @@ The following modules compose the standard library:
 
 *NOTE:* The txiki.js build depends on a number of git submodules ([libffi], [libuv] and [wasm3]).
 If you didn't already clone this repository recursively, make sure you initialize these
-submodules with `git submodule update --init` before proceeding to the build. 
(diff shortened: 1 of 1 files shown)


saghul:
I think it might be better to link with the system libffi TBH... I'm going to sleep a bit on this.

ahaoboy:
> I think it might be better to link with the system libffi TBH... I'm going to sleep a bit on this.

Have a good rest. I just found through searching that this compilation error can be fixed by installing libltdl-dev.

https://github.com/ahaoboy/js-engine-benchmark/actions/runs/11816168716/job/32918888844

PR #683, opened 2025-01-12
Initial commit for docs

KaruroChori:
Ok, I started adding everything needed to handle the docs.
- [x] tsdoc redirected
- [x] doxygen generated 
- [x] mkdocs generated
- [ ] branding (there is some stuff which is specific to the project I am adapting this code from which should be removed/edited)
- [ ] github worflows  

There are several styles available for mkdocs, we might replace it with something else.

Assuming you have all the dependencies installed, to test the generated docs:
```
./dist/make-docs.sh && cd dist/website && http-server
```
or something equivalent.

Dependencies:
```
python3 -m pip install mkdocs mkdocs-github-admonitions-plugin mkdocs-get-deps pymdown-extensions mdx_truly_sane_lists
```
and clearly `doxygen` must be installed as well.  
These dependencies are only if one wants to build the docs themselves, they will be handled automatically in the CI.


KaruroChori:
How did some platforms broke, as I have not even touched them I wonder.

saghul:
Thanks for getting the ball rolling! ❤️‍🔥

saghul:
A long time has passed, and somewhere there I decided to go with Docusaurus + a typedoc plugin: https://github.com/saghul/txiki.js/pull/786

PR #696, opened 2025-02-28
feat: add build flags

QuincySx:
Add a compilation flag to selectively remove non-essential modules, ensuring the build artifact remains compact.


saghul:
Sorry but I really don't want to do this piecemealing of features.

Issue #447, opened 2023-12-21
TextDecoding pretty slow


lal12:
I recently wondered why TextDecoding was so slow. Several seconds for 50-100 Kbytes (I am running on a low power embedded system with single core performance comparable to RPI 1).

I saw it currently uses a JS only implementation. I am wondering why, the `JS_NewStringLen` of quickjs already handles utf8 doesn't it?

As a workaround I am currently using 
```
import FFI from 'tjs:ffi';
export function quickDecodeUtf8(buf) {
    return FFI.bufferToString(buf);
}
```

Which also just does a `JS_NewString`.

saghul:
I think you might be right. Maybe running some WPT tests would help clarify if there are any missing corner cases.

saghul:
I remember we discussed that here https://github.com/saghul/txiki.js/issues/316 and didn't go for it, but I'm happy to revisit.

lal12:
> I remember we discussed that here #316 and didn't go for it, but I'm happy to revisit.

Ah, I misread that as only being about the changing of the JS polyfill.

lal12:
I copied (and adjusted) the TextDecoder tests from wpt.
The issue seems to be with streaming / multiple calls to TextDecoder. 
Which matches was described in #316.

lal12:
I will look into this tomorrow, trying to make a C implementation.

saghul:
Thanks for being so thorough!

As for Option 1, we do already use a fork, quickjs-ng, so we can discuss exposing what's necessary there.

saghul:
At this point it looks like having some tjs.encodeUTF8 / decodeUTF8 might the simplest way to go. Would that work for you? This would also simplify some of the internals I think, doing `new TextEncoder().endode(data)` gets old quick 😅

lal12:
> doing new TextEncoder().endode(data) gets old quick 😅

I agree, but I don't think I would be quite happy about that solution, since some JS libs just use TextDecoder internally. 
I think the more clean solution would be to have a standard conforming quick TextEncoder/TextDecoder. 

The qjs internals could just be copied into the Decoder module. And then  
- use minimal exposure of QJS internals (basically only allowing to create a string from external buffer)
- add an additional function which (if necessary) copies the string once, but doesn't check for any encoding stuff
- just letting the QJS internals as is, and doing the utf8 decode twice basically

saghul:
[...]

PR #700, opened 2025-04-30
add c utf8 encoder / decoder

lal12:
As discussed in #447 
A C implementation for TextEncoder/TextDecoder, due to the limited access to the QuickJS internals it has some unnecessary overhead (basically encoding/decoding is done twice), but it is still much faster than in JS.


lal12:
Some checks aren't succeeding due to #701

saghul:
Merged the other PR, can you rebase?

lal12:
@saghul any updates here?

saghul:
Alright, let's do this. Thanks for persevering @lal12 ❤️

PR #708, opened 2025-05-20
core:fix missing accept in connection_cb

partic2:


saghul:
What are you fixing here? Can you create a test case?

partic2:
I'm trying to implementing a http server on txiki.js. 
When I use Firefox(on Windows) to test the server, Some connections worked fine,  Some connections didn't response any data(also didn't refused). 
After checking the code,  I guess the cause is that tjs doesn't accept socket if there is NO pending accept call from JS side. 
https://github.com/saghul/txiki.js/blob/6e01bf4a0cbc1aba6dcdddf1d0aa67df287ce420/src/mod_streams.c#L341-L344

And github copilot tell me
> "Do NOT call uv_accept outside the connection callback (uv_connection_cb)." 

So this patch queue socket to avoid socket lost.

saghul:
It must be something else, that advice is wrong.

See: https://github.com/libuv/libuv/blob/71ec5c0fcdd867b64c46ade1e0a6b59101281a4a/src/unix/stream.c#L528

saghul:
That said, I do have plans to change the socket API to align it with Chrome's direct sockets API and WinterTC's one, so I'll take a look then.

partic2:
I'll show you the different.

Below code is a simple echo server, expected to print 'hello' 
But it will print nothing on current version txiki.js,
And it will print "hello" on this patch.

```javascript
async function test() {
    let serv = await tjs.listen('tcp', '127.0.0.1', 1180);
    (async () => {
        const conn = await tjs.connect('tcp', '127.0.0.1', 1180);
        await conn.write(new TextEncoder().encode('hello'));
        let msg = new Uint8Array(10);
        msg = msg.subarray(0, (await conn.read(msg)));
        console.info(new TextDecoder().decode(msg));
    })();
    await new Promise((resolve)=>setTimeout(resolve, 1000));
    while (true) {
        const conn = await serv.accept();
        let msg = new Uint8Array(10);
        msg = msg.subarray(0, (await conn.read(msg)));
        await conn.write(msg);
    }
}
test()
```

saghul:
I changed the entire sockets approach in https://github.com/saghul/txiki.js/pull/773 so this no longer applies, sorry!

PR #709, opened 2025-05-20
polyfills:make event.target refer to event.currentTarget

partic2:


saghul (approved):

partic2:
To avoid unnecessary change on bundles ".c“ file, which may cause a significant expansion of the repository's size. I'd prefer to generate bytecode once after https://github.com/saghul/txiki.js/pull/710 is also accepted. 
If you think it's better to update bytecode for every commit, I'll regenerate it for each commit later.

saghul:
The bytecode needs to be updated on each PR.

PR #721, opened 2025-08-26
feat: implement XMLHttpRequest.withCredentials

### LINKED ISSUE #720: Typo in XHR.withCredentials
It seems there is a typo in JS binding of `XMLHttpRequest.withCredentials`:
https://github.com/saghul/txiki.js/blob/feaedc337db171f687c8c84b52c95ed56b804edd/src/js/polyfills/xhr.js#L103-L109
Both in C and fetch polyfill use `withCredentials` insteadof `withCcredentials`
https://github.com/saghul/txiki.js/blob/feaedc337db171f687c8c84b52c95ed56b804edd/src/xhr.c#L815
https://github.com/saghul/txiki.js/blob/feaedc337db171f687c8c84b52c95ed56b804edd/src/js/polyfills/fetch/fetch.js#L53-L57

KawaiiZapic:
Implemented XHR.withCredentials with cURL's cookies jar
Also fixed #720


KawaiiZapic:
> Could you tame the linter please?

Sorry for that, I have run the linter for C and JS

KawaiiZapic:
> Could you tame the linter please?

Maybe there is some difference of clang-format between Homebrew version and Linux version, I have re-formatted it manually, according the output log of workflow

saghul:
The test failures seem related.

KawaiiZapic:
> The test failures seem related.

It seems some limitations of GitHub Action are preventing cURL from saving cookies to `$HOME/.tjs`, I downloaded [the binary that failed to pass tests](https://github.com/KawaiiZapic/txiki.js/actions/runs/17294193686) and it just works fine on my machine, all tests are passed.  
   
I can't figure out why this is happening, so I make `$TJS_HOME` point to `/tmp` when testing cookies, and [it works fine](https://github.com/KawaiiZapic/txiki.js/actions/runs/17294696947/job/49089930764).
https://github.com/KawaiiZapic/txiki.js/blob/4a11f0e5be186b2fca6a9e09ae28959357370a56/tests/test-xhr-credentials.js#L40-L48  
I don't think this is the best solution for this problem. Is this acceptable?

saghul:
Perhaps use spawn to run the test in a nested environment? Or use `tjs.homeDir` as the jar path?

src/js/polyfills/xhr.js
```diff
@@ -106,14 +106,15 @@ class XMLHttpRequest extends EventTarget {
             const path = globalThis[Symbol.for('tjs.internal.modules.path')];
             const TJS_HOME = tjs.env.TJS_HOME ?? path.join(tjs.homeDir, '.tjs');
 
-            this[kXHR].cookieJar = path.join(TJS_HOME, 'cookies');
+            tjs.makeDir(TJS_HOME, { recursive: true });
```
saghul, line 109: This returns a promise, so it's not guaranteed to finish before your next statement. Plus, if feels wrong to do here.

saghul: There is an internal mkdirSync function in storage.js you may want to move that one around to use it.

tests/test-xhr-credentials.js
```diff
@@ -1,4 +1,11 @@
 import assert from 'tjs:assert';
+import path from "tjs:path";
+
+// Cookies jar must be pre-created, otherwise this test will failed in Github Action
+const TJS_HOME = tjs.env.TJS_HOME ?? path.join(tjs.homeDir, '.tjs');
+const cookieJarPath = path.join(TJS_HOME, 'cookies');
+(await tjs.makeDir(TJS_HOME, { recursive: true }));
+(await tjs.open(cookieJarPath, 'w')).close();
```
saghul, line 8: This shouldn't be necessary, the code should be able to work without an existing .tjs directory. See my comment above.

saghul (commented):
Almost there!

src/js/polyfills/xhr.js
```diff
@@ -100,12 +102,20 @@ class XMLHttpRequest extends EventTarget {
...
     }
 
-    set withCcredentials(value) {
-        this[kXHR].withCcredentials = value;
+    set withCredentials(value) {
+        if (value) {
+            const path = globalThis[Symbol.for('tjs.internal.modules.path')];
+            const TJS_HOME = tjs.env.TJS_HOME ?? path.join(tjs.homeDir, '.tjs');
+
+            mkdirSync(TJS_HOME, { recursive: true });
```
saghul, line 110: Can you perhaps add a flag so we only do this initialization once?

saghul: Agreed, let's do that afterwards 👍

saghul:
Cheers!

PR #723, opened 2025-09-06
Fix/fetch content type blob

### LINKED ISSUE #722: fetch() does not set MIME type in .blob() response
When using fetch, the content type of the blob is not set. As an example you can run the sample below.
`b.type` is going to be an empty string.

```js
fetch('https://picsum.photos/200')
        .then(res => res.blob())
        .then(b => {
            console.log('Content type:', b.type);
        })
        .catch(console.error)
```

That's because in the fetch polyfill (/src/js/polyfills/fetch/body.js) the blob is not set with the type option.

The fix would be to just take the `Content-Type` header value and set it as the type.

Throvn:
This PR should close #722.

It added a test (`fetchWithBlobBody`) to verify that the blob type is set correctly.

It also set the Blob type to the content type header of the fetch response. If the content type is not defined, it falls back to an empty string (like before).


saghul:
I think you need to run make js again

Throvn:
Thanks!

```
throvn@Louiss-MBP-2 txiki.js % git branch
* fix/fetch-content-type-blob
  master
throvn@Louiss-MBP-2 txiki.js % make js
cmake --build build --target tjsc -j 8
gmake[1]: Entering directory '/Users/throvn/Documents/txiki.js/build'
[ 75%] Built target qjs
[100%] Built target tjsc
gmake[1]: Leaving directory '/Users/throvn/Documents/txiki.js/build'
throvn@Louiss-MBP-2 txiki.js % git diff
throvn@Louiss-MBP-2 txiki.js % 
```

`make js` doesn't seem to change anything for me.

PR #734, opened 2026-01-07
feat: implement `redirect` property for fetch

jspngh:
This implements the `redirect` functionality for the Fetch API.
More info on the redirect property [can be found here](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#redirect).

Since fetch is implemented in terms of `XMLHttpRequest`, this poses some issues.
Standard `XMLHttpRequest` always follows redirects and does not provide an API to change the behaviour.
Therefore, I had to add a non-standard `redirectMode` property.

I do believe this is justified, since the Fetch API is promoted as a replacement for `XMLHttpRequest` (e.g. [here](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest_API)).
So having a more feature-complete implementation for Fetch should take priority, in my opinion.

Internally it is implemented by:
- setting [`CURLOPT_FOLLOWLOCATION`](https://curl.se/libcurl/c/CURLOPT_FOLLOWLOCATION.html) to 0 in case of `'manual'` redirects
- setting [`CURLOPT_MAXREDIRS`](https://curl.se/libcurl/c/CURLOPT_MAXREDIRS.html) to 0 in case of `'error'` redirects.

What do you think?

### DIFF
--- a/tests/test-xhr-progress.js
+++ b/tests/test-xhr-progress.js
@@ -2,7 +2,7 @@ import assert from 'tjs:assert';
 
 
 let lengthComputable = false,  loaded, total;
-const url = 'https://postman-echo.com/get';
+const url = 'https://httpbin.org/get';
 const xhrSync = new XMLHttpRequest();
 xhrSync.open('GET', url, false);
 xhrSync.onprogress = (evt) => {


saghul:
Hum, I'll take a look.

tests/test-xhr-redirect.js
```diff
@@ -0,0 +1,34 @@
+import assert from 'tjs:assert';
```
saghul: Can you please write the test using fetch instead? The underlying implementation may change and not be exposed in XHR eventually, but fetch should work.

PR #737, opened 2026-01-12
fix: prevent premature garbage collection of xhr in fetch polyfill

### LINKED ISSUE #736: Fetch sometimes never resolves
I ran into a tricky bug where the combination of opening a websocket and calling fetch at about the same time
caused the fetch call to be 'stuck', i.e. it didn't do anything.

After some trail and error I finally figured out it was because `tjs_xhr_finalizer` was called right after doing the fetch call.
At this point I was a bit out of my depth, so I asked AI to chime in (that is to say: take the explanation and proposed solution with a grain of salt).

The theory is that due to the websocket activity, garbage collection is triggered earlier/more than in the case when fetch is called on its own.
The `XMLHttpRequest` that is created in the Promise for `fetch` does not have any strong reference to it and gets cleaned up by quickjs.

I also asked it for a more minimal reproducer and a fix, which you can have [a look at here](https://github.com/saghul/txiki.js/compare/master...jspngh:txiki.js:bugfix/fetch-gc).

The reproducer starts a fetch that will take some seconds to complete and then triggers garbage collection at a high frequency.
Additionally I also added a print statement in `tjs_xhr_finalizer`.
The behaviour I see here is the same: `tjs_xhr_finalizer` is called quickly after starting and the fetch call never completes.

The proposed fix is to add an `activeXHRs Set` with references to `XMLHttpRequest`s that are in progress.
When I include this change, the reproducer also behave correctly (the fetch completes and `tjs_xhr_finalizer` is only called after completion).

I'm reasonably confident that this is indeed the root cause, but I will leave the evaluation of the solution up to you.

jspngh:
The `XMLHttpRequest` that is created as part of the `fetch` call could be prematurely garbage-collected because no strong reference to it exists anymore.

When this happens (garbage collections runs while the fetch is in progress), this leads to a promise that will never resolve.

To fix this, we add a `Set` that tracks all the active XHR's.

This closes #736

### DIFF
--- a/tests/test-fetch-gc-regression.js
+++ b/tests/test-fetch-gc-regression.js
@@ -1,0 +1,22 @@
+import assert from 'tjs:assert';
+
+// This test simulate aggressive GC that might happen with concurrent operations.
+// It aims to detect regressions in the `fetch` implementation.
+const gcInterval = setInterval(() => {
+    tjs.engine.gc.run();
+}, 50);
+
+Promise.race([
+    fetch('https://postman-echo.com/delay/2'),
+    new Promise((resolve, reject) => {
+        setTimeout(() => reject(new Error("XHR has been garbage collected")), 3000);
+    }),
+])
+    .then((res) => {
+        assert.eq(res.status, 200);
+        clearInterval(gcInterval);
+    })
+    .catch((err) => {
+        assert.ok(false, err);
+    })
+


saghul (commented):
Please commit the result of `make js` and add the repro test to the tests dir so it's tested going forward.

PR #739, opened 2026-01-20
Featture: expose stringToBuffer in ffi

ldeninski:
This PR exposes a stringToBuffer function that is the inverse of bufferToString. This is useful for writing utf-8 encoded strings to sockets. It is intended to be used along with the changes in https://github.com/KawaiiZapic/h3-txikijs-adapter/pull/1

A logical next step would be to implement a proper TextEncoder and TextDecoder that are using these functions as the js polyfills are really inefficient.

P.S. I will try to make the time for a proper decoder and encoder implementation if the maintainers are okay with that.

### DIFF
--- a/types/src/ffi.d.ts
+++ b/types/src/ffi.d.ts
@@ -87,7 +87,7 @@ declare module 'tjs:ffi'{
     }
 
     export function bufferToString(buf: Uint8Array): string;
-    export function stringToBuffer(s: string): Uint8Array<ArrayBuffer>;
+    export function stringToBuffer(s: string): Uint8Array;
 
     export class Pointer<T, N extends number>{
         constructor(addr: PointerAddr, level: N, type: SimpleType<T>);


ldeninski:
> > P.S. I will try to make the time for a proper decoder and encoder implementation if the maintainers are okay with that.
> 
> There is a PR #700

@lal12, can you push some small change so CI can recheck. Some windows build is failing and the logs have expired. Also would you be okay if I suggest some cosmetic changes? I reviewed the code with my limited c experience and find couple of functions hard to read due to missing defines/constants in returns. I can add some comments to the PR if you are okay with that.

P.S. I am by no means experienced in c/c++ or affiliated with this project :)

lal12:
@ldeninski I rebased the PR. So let's see if it goes through the CI...
But even if not, this should be a smaller issue, so you can still comment on it.

Btw. I tried to follow the specs pseudo code as closely as possible, so hopefully spec compliant, but not the most efficient way. Also it has significant overhead, because of QucikJS internals (see #447)

PR #770, opened 2026-02-08
WebStreams all the things

saghul:
- [x] child process stdio
- [x] own process stdio
- [x] sockets
- [x] files (optional, with readable / writable properties)


saghul:
I'm considering replacing console.log's implementation with just fwrite + fflush. It has its drawbacks, but maybe it's for the best...

Can you give me a WASM example?

saghul:
Can you give https://github.com/saghul/txiki.js/pull/774 a try?

ldeninski:
> Can you give #774 a try?

That looks correct now :)
```
16:06 $ ../txiki.js/build/tjs run src/examples/test.js 
console.log before print_from_wasm
Print from WASM
console.log after print_from_wasm
```

PR #773, opened 2026-02-10
socket: implement the Direct Sockets API

saghul:


examples/echo-server-pipe.js
```diff
@@ -5,17 +5,25 @@
...
 
-let conn;
 while (true) {
-    conn = await p.accept();
+    const { value: conn, done } = await reader.read();
+
+    if (done) {
+        break;
+    }
+
```
ldeninski: Will this work?

```js
for await (const conn of readable) {
    handleConnection(conn)
}
```

PR #784, opened 2026-02-17
http: add HTTP server

saghul:


fayakun-it-consulting:
Any way to set TLS keys?

PR #814, opened 2026-03-03
build: improve Makefile macOS compatibility

adam-ikari:
Improve Makefile macOS compatibility.

## Changes

- Use single-line JOBS detection with proper fallback chain (getconf → nproc → sysctl)
- Improve `distclean` to delete all generated files in bundles directory while preserving directory structure

## Details

The original JOBS detection caused Make to exit with an error on some macOS systems. This happened because separate shell calls with `ifeq` checks could return non-zero exit codes, which Make treats as fatal errors.

The new implementation uses a single shell command chain with `||` operators, ensuring:
- All fallback commands are attempted in sequence
- The command chain always succeeds (final `echo 4` guarantees a return value)
- No error propagation to Make that would cause build failures

This fix resolves the issue where `make` would immediately exit on certain macOS configurations.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,9 +2,6 @@ BUILD_DIR=build
 BUILDTYPE?=Release
 
 JOBS?=$(shell getconf _NPROCESSORS_ONLN 2>/dev/null || nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null || echo 4)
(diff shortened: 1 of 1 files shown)


saghul:
Can you give an example of when the jobs check would fail?

PR #909, opened 2026-04-20
building.md : +hint for which Ubuntu version compiles/fails

jerng:


saghul:
Can you move this to the GNU/Linux section instead, and say something like "Note: If using Ubuntu, version >= 24.04 required)".

saghul (approved):

PR #923, opened 2026-05-17
[docs] Update code-bundling.md

houd1ni:
- Added `Other esbuild options` section.
- Removed `use esbuild directly` as it seems that `tjs bundle` covers all possible CLI cases.

while developing [txiki-utils](https://codeberg.org/houd1ni/txiki-utils) I have used esbuild directly, but as @saghul [noticed](https://github.com/saghul/txiki.js/discussions/922#discussioncomment-16928698), it could be done via [tjs bundle](https://txikijs.org/docs/guides/code-bundling). But until I had a look at [the sources](https://github.com/saghul/txiki.js/blob/f47617eb37830941a5a70ee572ce034d85bf7d8a/src/js/run-main/bundle.js#L150), I could not notice that it is possible to pass esbuild CLI parameters in `tjs bundle`.

Also while I was editing the docs to add the section, I finally got that [the very last one](https://txikijs.org/docs/guides/code-bundling#using-esbuild-directly) with its `There are other interesting options available` is exactly about that, so I've decided to remove it as it seems that `tjs bundle` already covers all possible CLI cases indeed. But I would not argue if it would stay there (:

### DIFF
--- a/website/docs/guides/code-bundling.md
+++ b/website/docs/guides/code-bundling.md
@@ -44,6 +44,14 @@ tjs bundle --minify my-app/index.ts bundle.js
 
 This is especially useful when creating [standalone executables](standalone-executables.md), as the bundled code is embedded in the resulting binary.
 
+### Other esbuild options
+
+Any other cli option of esbuild is also available in `tjs bundle`:
+
+```bash
+tjs bundle --minify --define:DEBUG=true --drop:debugger my-app/index.ts
+```
+
 ### How it works
 
 The first time you run `tjs bundle`, the esbuild binary is downloaded from the npm registry and cached in `~/.tjs/esbuild/<version>/`. Subsequent runs reuse the cached binary. The download requires no system tools since it uses the built-in `fetch` API and `DecompressionStream` to stream the tarball directly to disk.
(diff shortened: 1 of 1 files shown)


website/docs/guides/code-bundling.md
```diff
@@ -97,20 +105,3 @@ Error: something went wrong
...
-    --bundle \
-    --outfile=bundle.js \
-    --external:tjs:* \
-    --target=esnext \
-    --platform=neutral \
-    --format=esm \
-    --main-fields=main,module
-```
-
-There are other interesting options available, check the [API documentation](https://esbuild.github.io/api/).
```
saghul, line 116: Please keep this sentence and link in the new section you added.

website/docs/guides/code-bundling.md
```diff
@@ -97,20 +105,3 @@ Error: something went wrong
 ```
 
 Both inline source maps (`data:` URLs) and external `.map` files are supported.
-
-## Using esbuild directly
```
saghul, line 101: I think we can keep this, but maybe change the part that says if you want more control voer the process.

Maybe someone wants to create a bundle for use with tjs but can't use tjs to produce it, somehow?

website/docs/guides/code-bundling.md
houd1ni: done

website/docs/guides/code-bundling.md
houd1ni: if they already have it as a dep and don't want to pull another one maybe... anyway, it's in the bottom if nothing else helped. revived.

website/docs/guides/code-bundling.md
houd1ni: maybe `for the full reference take a look at the [API documentation](https://esbuild.github.io/api/).` there ?

saghul (approved):

PR #955, opened 2026-06-04
Fix macOS plugin build issues

lpla:
Fix two macOS build issues hit when txiki is embedded in RetroPlug AU builds.

- Keep the POSIX socket module source private to the `tjs` target so it is not pulled into other linked targets.
- Select WAMR's Darwin target from single-arch `CMAKE_OSX_ARCHITECTURES`, so an arm64-only build uses `AARCH64` instead of x86_64 assembly.

Validation:
- Built RetroPlug AU on macOS for x86_64.
- Built RetroPlug AU on macOS for arm64 using host-tool overrides.
- Built RetroPlug AU as universal `x86_64;arm64`; `lipo` confirmed both slices and `auval -v aumu RPlg Dstr` passed on the Intel host.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,7 +126,7 @@ add_library(tjs STATIC
 )
 
 if(UNIX)
-    target_sources(tjs PUBLIC src/mod_posix-socket.c)
+    target_sources(tjs PRIVATE src/mod_posix-socket.c)
 endif()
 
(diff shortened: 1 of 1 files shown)


saghul (approved):

saghul:
Thank you! TIL about RetroPlug, it's nice to see txiki.js was useful there!

Issue #911, opened 2026-04-29
typescript for core js ?


houd1ni:
hi! the project is **awesome**, thanx!

I wonder why is the https://github.com/saghul/txiki.js/tree/master/src/js part written in plain js, not ts other than that it's kinda native and more predictable in some sense ?

saghul:
Thanks for the kind words!

I don't remember exactly how I landed there.

Might have been because I wasn't using esbuild from the get go, so there was no bundling step.

At this point I just don't feel like rewriting it to TS, as the benefit won't be that big, but I may add new code in TS.

houd1ni:
Oh, it's a surface to improve then, PRs are welcome?)

saghul:
Sure thing!

PR #958, opened 2026-06-04
ffi: add zero-copy pointer views

saghul:
Provide a way to get a Uint8Array/ArrayBuffer that aliases native
memory without copying, plus a way to safely invalidate such a view.

- NativePointer.toUint8Array()/toArrayBuffer() methods and matching
  toUint8Array()/toArrayBuffer() free functions. The free functions
  take an `owner` option that pins the JS object owning the memory so
  the GC can't reclaim it while a view is alive.
- detachBuffer() neutralizes a view after the native memory is freed,
  turning a potential use-after-free into a harmless empty buffer
  without reading the bytes (unlike ArrayBuffer.prototype.transfer()).


houd1ni:
dammit, that works! just packed some data using brotli and vise-versa!

by the way, while I am using more of this,  why there are two APIs for that ? I mean free and method fns. if the owner prop is needed, it could be added as one more arg if I got this right. should be easier to learn and maintain.

kinda this (remix from [there](https://github.com/saghul/txiki.js/blob/5d919fc523e813633e5155b3bc049b8e57fd16d0/website/docs/guides/ffi.md#working-with-native-memory-zero-copy)) should already work and does not require free func:
```typescript
import type { NativePointer } from 'tjs:ffi';

class Painting {
  view: Uint8Array
  tuneRed(pixel: number, color: number) { this.view[pixel*4] = color%256 }
  constructor(
    private ptr: NativePointer,
    private w: number,
    private h: number
  ) { this.view = ptr.toUint8Array(w * h * 4) }
}
const modern_art = new Painting(symbols.get_pixels(), 1, 1)
modern_art.tuneRed(0, 255)
// now easily eBay it for €1M.
```

and remix of [this](https://github.com/saghul/txiki.js/blob/5d919fc523e813633e5155b3bc049b8e57fd16d0/website/docs/guides/ffi.md#lifetimes) should be more straightforward especially after [Node APIs](https://nodejs.org/api/buffer.html#static-method-bufferfromarray):
```typescript
import { NativePointer, read } from 'tjs:ffi';
import { ok } from 'tjs:assert'

const src = new Uint8Array([1, 2, 3, 4]);
const view = toUint8Array(NativePointer.fromBuffer(src), src.length, 0); // may it carry src as an owner here ?

// more step-by-step (this looks extremely intuitive for me as for js/ts dev):
const src = new Uint8Array([1, 2, 3, 4])
const ptr = NativePointer.fromBuffer(src) // owner taken. are there any cases where it would be undesirable ?
const view = ptr.toUint8Array(/*from?, to?*/) // might be even Array.prototype.slice() api ?.. though in the class api below it doesn't look so cool mainly because offset is used more rarely than length.
// and when the view is collected, all its' chain got collected too, calling detachBuffer() inside.

// hence this might be super-duper nice api (part of the code):
class NativePointer {
  static fromBuffer(src: Uint8Array, takeOwner = true) { return new NativePointer(..., src) }
  public toUint8Array(length: number, offset = 0) { /* calls C func as it already does. */  }
  public dispose() { ... if(this.src) detachBuffer(this.src) ... }
  [Symbol.dispose]() { this.dispose() }
  // and what's about adding read stuff here too ?.. thinking just by the way. ))
  public u8(offset = 0) { return read.u8(this, offset) }
  constructor(..., private src?: TypedArray) { // TypedArray type is taken from somewhere.
    ok(src===undefined || ArrayBuffer.isView(src)) // if really needed.
  }
}
```

PS: lol, I'm sort of started https://github.com/saghul/txiki.js/issues/911#event-25041649188 here 😆

saghul:
I made another round. Dropped the free-functions. THe Uint8Array / ArrayBuffer that is returned from the APIs now are instances of `ExternalArrayBuffer` which have an extra function in their prototype: `detach()`.

I think it feels cleaner now.

The one thing I haven't tackled (and I'm not sure I should) is auto-detaching views _somehow_. An option would be to latch the views onto the pointer instance, and when the pointer object goes away they'd be detached automagically. This becomes weird with offset() / deref() since they return fresh objects.

For now I think it's ok to stop here, since this last thing can be added afterwards, once we have used this more. That's what Deno currently does FWIW.

houd1ni:
am trying to use all of those, will take a bit.

PR #962, opened 2026-06-05
Build: tjs-lite - customize build: disable features (WASM, SQLite, TLS)

lukasMega:
Discussion: https://github.com/saghul/txiki.js/discussions/954

# What this is: 

Makes WASM, SQLite, and TLS optional at build time (`BUILD_WITH_*=OFF`), plus size knobs (`STRIP/LTO/GC_SECTIONS/MinSizeRel`). 
Adds a runtime `tjs.engine.features` object and feature-gated test skipping. 
**~6.1 MB → ~2.1 MB fully stripped.**

## Suggested read order (≈5 logical groups)
  
1. Build switches & guards — `CMakeLists.txt`. Confirm each flag sets its `TJS_NO_*` define and gates the right target_sources / target_link_libraries. Note: when TLS is off, lws builds without SSL and only mbedcrypto is linked (so WebCrypto survives).

2. C feature gates — `private.h` (struct fields, headers, decls), `vm.c` (init/cleanup), `builtins.c` + `mod_engine.c` (tjs.engine.features object). Check every gated symbol is gated consistently across decl/init/use.

3. C runtime guards — `httpclient.c`, `httpserver.c`, `ws.c`, `lws-utils.c`. These reject https/wss/cert-bearing servers with a clear error when TLS is off (no silent plaintext downgrade) — verify the error paths free everything.

4. JS layer — `engine.js` (exposes features), `polyfills/wasm.js` + `direct-sockets/tls.js` (guard global/constructors), `run-main/index.js` + `run-repl/history.js` (static→lazy import() of `tjs:wasi/tjs:sqlite`), and `run-main/skip.js` + `tests/feature-skip.json` (single-* glob skipper).
  

## Build options (all default ON, independent and composable):
- `BUILD_WITH_WASM=OFF`    drop WebAssembly/WAMR + WASI (~0.4 MB)
- `BUILD_WITH_SQLITE=OFF`  drop SQLite + REPL history (~1.6 MB)
- `BUILD_WITH_TLS=OFF`     drop HTTPS/WSS/tjs:tls; WebCrypto stays (~0.7 MB)
Plus size knobs: `BUILD_WITH_STRIP`, `BUILD_WITH_LTO`, `BUILD_WITH_GC_SECTIONS`, and `MinSizeRel` support.

---

Build/test matrix to run

```bash
  make && ./build/tjs test tests/                               # default: all features on
  cmake -B b-notls  -DBUILD_WITH_TLS=OFF    && cmake --build b-notls  -j
  cmake -B b-nowasm -DBUILD_WITH_WASM=OFF   && cmake --build b-nowasm -j
  cmake -B b-nosql  -DBUILD_WITH_SQLITE=OFF && cmake --build b-nosql  -j
  for b in b-notls b-nowasm b-nosql; do ./$b/tjs test tests/; done   # feature tests should SKIP, rest pass
  ./b-notls/tjs eval "console.log(JSON.stringify(tjs.engine.features))"   # {wasm,sqlite,tls:false}
```

Already verified locally: TLS-off compiles warning-free, HTTPS throws cleanly, TLS tests SKIP, test-webcrypto-rsa.js passes. Suggest adding a CI matrix entry per flag so the `#ifdef` paths don't bit-rot.


saghul:
@lukasMega I took a closer look, there is some good work here!

Let's split it in different parts and work on merging them individually. I suggest these:

- Build flags for stripping and LTO
- Build flag for no SQLite
- Build flag for no WASM

Please check the current "no ffi" flag and make the new flags mirror it.

You may notice I left TLS out. I'd like to sit on that a bit longer since it's more invasive. We lose TLS sockets, but also the hashing module and the entirely of WebCrypto, so that might require some extra consideration.

I deduplicated cacert on master (great catch!) but didn't apply compression. Compression can also be considered for the bundled bytecode, at the cost of a bit of startup time. Maybe some custom builds are happy with that tradeoff, I'd accept a PR making it optional, defaulting to OFF.

Last, I looked a bit into the flags we use for SQLite, and there seems to be some low hanging fruit we might be able to collect: FTS3 is builtin, but FTS5 is the successor, so we can save ~175K by removing FTS3, I'll take a look at that since it seems like free gains.

saghul:
I'm going to close this since all its parts are either merged or in PR. Nice work here @lukasMega !

PR #970, opened 2026-06-11
build: support cross-compilation for Android (bionic)

loss-and-quick:
# Summary

Adds the necessary build-system changes to cross-compile txiki.js for
Android (arm64-v8a, x86_64) using the Android NDK, and fixes a runtime crash on
Android caused by lws's async DNS resolver.

# Changes

**`BUILD_WITH_FFI` option** — The `tjs:ffi` module depends on libffi, which is
not available in Android NDK sysroots. The new option (default ON) wraps
`mod_ffi.c` and its libffi linkage behind `TJS_HAVE_FFI`, so it can be turned
OFF for cross-builds:

```bash
cmake -DCMAKE_TOOLCHAIN_FILE=$NDK/build/cmake/android.toolchain.cmake \
      -DANDROID_ABI=arm64-v8a \
      -DBUILD_WITH_FFI=OFF \
      ../..
```

When `BUILD_WITH_FFI=OFF`, the C sources are excluded from the build,
`mod_ffi.c` is not compiled, and the `tjs:ffi` builtin module is not
registered.

**WAMR platform selection** — On Android, WAMR is built with
`WAMR_BUILD_PLATFORM=android` instead of the default `linux`. The linux
platform pulls `<sys/timeb.h>` which bionic does not provide.

**WAMR target architecture** — When `ANDROID_ABI` is set to `arm64-v8a` or
`x86_64`, `WAMR_BUILD_TARGET` is set to `AARCH64` / `X86_64` automatically.

**liblog linkage** — WAMR's android platform logs via `__android_log_*` from
bionic liblog, so `vmlib` links `-llog` on Android.

**lws async DNS fix** — Modern Android (8+) has neither `/etc/resolv.conf` nor
populated `net.dns1..4` system properties. lws's async DNS resolver
(`LWS_WITH_SYS_ASYNC_DNS`) crashes with SIGSEGV when the nameserver list is
empty.  So we are disable it on Android so the client connect path falls back to the
blocking OS resolver (`getaddrinfo`, served by netd). This makes `fetch()` work
correctly on Android devices.

# Files changed

- `CMakeLists.txt` — `BUILD_WITH_FFI` option, conditional `mod_ffi.c`
  source/link, WAMR platform/target for Android, liblog link, conditional
  `LWS_WITH_SYS_ASYNC_DNS`
- `src/builtins.c` — guard `tjs:ffi` bundle with `#ifdef TJS_HAVE_FFI`
- `src/vm.c` — guard `tjs__mod_ffi_init` with `#ifdef TJS_HAVE_FFI`


saghul (commented):
Is it necessary to disable FFI on Android?

Also, this would need a CI to guarantee it doesn't get broken in the future.

saghul:
Why did you need to disable async DNS?

loss-and-quick:
> Is it necessary to disable FFI on Android?

 The NDK sysroots don't ship libffi — there's no `ffi.h` and no `libffi.a`/`.so` in the prebuilt toolchain, so `find_library(ffi REQUIRED)` fails at configure time and the `tjs:ffi` module can't link on Android.

`BUILD_WITH_FFI` defaults to `ON`, so no other platform is affected — it only lets Android opt out of the one module that can't be satisfied by the NDK. The alternative is to vendor/build libffi for each Android ABI and keep FFI enabled; I'm happy to go that route instead if FFI-on-Android is something you'd like to support, it just adds a prebuilt dependency per ABI. Let me know which you prefer.

---

> Why did you need to disable async DNS?

On modern Android (API 26+) there's no `/etc/resolv.conf`, and the legacy `net.dnsN` system properties that `LWS_PLAT_ANDROID`/`android-resolv.c` reads are empty — DNS is resolved by `netd`. So lws's `LWS_WITH_SYS_ASYNC_DNS` resolver comes up with an empty nameserver list and crashes in `lws_async_dns_query` (NULL deref → SIGSEGV). I confirmed this on Android 16/arm64; the backtrace bottoms out at `lws_async_dns_query` ← `lws_client_connect_2_dnsreq`.

Disabling `LWS_WITH_SYS_ASYNC_DNS` on Android makes the client-connect path fall back to the blocking `getaddrinfo` (`lws_getaddrinfo46`), which is served by `netd` and resolves correctly on-device — verified with http/https, raw IPs and hostnames. Other platforms are untouched. I can attach the full backtrace and a minimal repro if that's useful.

---

> Also, this would need a CI to guarantee it doesn't get broken in the future.

I'll add a CI job that cross-compiles for Android (arm64 via the NDK toolchain) so this path doesn't regress. Will push it to this PR.

saghul:
Thanks, makes sense.

saghul (approved):

PR #971, opened 2026-06-11
lws: support SOCKS5 proxy

loss-and-quick:
# Summary

Add SOCKS5 proxy support to libwebsockets-based HTTP/WebSocket client
connections via environment variables, matching the existing HTTP CONNECT
proxy workflow.

`LWS_WITH_SOCKS5` — enabled in `CMakeLists.txt` so libwebsockets is built
with SOCKS5 client support.

`tjs__parse_proxy_url()` now recognizes `socks5://`, `socks5h://` (remote
DNS variant), and `socks://` schemes. These set `is_socks5 = true` on the
proxy config; everything else is treated as HTTP CONNECT as before.

`tjs__create_client_vhost()` routes SOCKS5 URLs through
`lws_context_creation_info.socks_proxy_address` / `socks_proxy_port`
instead of `http_proxy_address`. The unused proxy field is pinned to `""`
so lws does not fall back to reading its own env vars (`http_proxy` /
`socks_proxy`).

`tjs__lws_load_http_once` — same SOCKS5 routing for the one-shot HTTP load
path used by `import()` and `fetch()` internals.

`proxy_configs_equal()` includes `is_socks5` in the comparison to prevent
collapsing identical HTTP and SOCKS5 proxy configs into one vhost.

`no_proxy` works identically for both proxy types.

```bash
# SOCKS5 with local DNS
http_proxy=socks5://10.0.0.1:1080 ./build/tjs run app.js

# SOCKS5 with remote DNS (resolved by proxy)
http_proxy=socks5h://proxy:1080 ./build/tjs run app.js

# all_proxy fallback
all_proxy=socks5://proxy:1080 ./build/tjs run app.js

# no_proxy exclusion
no_proxy=localhost,127.0.0.1 http_proxy=socks5://proxy:1080 ...
```

# Files changed
`CMakeLists.txt`, `src/lws-utils.c`


src/lws-utils.c
```diff
@@ -649,9 +672,24 @@ static int tjs__lws_load_http_once(TJSRuntime *qrt, TJSHttpLoadCtx *load_ctx, co
...
+    } else if (have_proxy) {
         info.http_proxy_address = proxy_cfg.auth_address;
         info.http_proxy_port = proxy_cfg.port;
+#if defined(LWS_WITH_SOCKS5)
+        info.socks_proxy_address = "";
+#endif
+    } else {
+        /* No proxy: pin both so lws does not read http_proxy/socks_proxy env. */
+        info.http_proxy_address = "";
+#if defined(LWS_WITH_SOCKS5)
```
saghul: Can you please drop all these checks? Since we use the embedded static build it's guaranteed to be there.

loss-and-quick:
The test `tests/test-xhr-credentials.js` is a cookie jar integration test: it sets a cookie via `Set-Cookie` response header and expects it to be sent back on the next request with `withCredentials: true`.

`lws_cookie_attach_cookies()` in the vendored libwebsockets returns `-1` when `wsi->stash` is NULL, without attempting to read the same data from the wsi's HTTP headers. The stash is freed early during client connect initialisation because the txiki.js build enables `LWS_WITH_SOCKS5` (the stash-free logic is guarded by `#if defined(LWS_WITH_SOCKS5)` in `connect.c`).

The sister function `lws_cookie_write_nsc()` — which handles saving cookies from `Set-Cookie` — already has a fallback via `lws_hdr_simple_ptr()`. But `lws_cookie_attach_cookies()` did not — hence cookies were saved but never sent.

Track in: https://github.com/warmcat/libwebsockets/pull/3608

saghul:
Cool! Let's keep this PR open but blocked on that lws PR.

loss-and-quick:
> Cool! Let's keep this PR open but blocked on that lws PR.

FWIW, I've now cursed libwebsocket about three times trying to integrate txiki.js into my project

loss-and-quick:
@saghul Fixed in upstream libwebsocket

saghul:
I landed the lws update. Can you please rebase?

saghul:
There seems to be a build error in lws on windows.

loss-and-quick:
> There seems to be a build error in lws on windows.

Not my fault, libwebsocket doesn't have guard for MSVC's C4267: casting (`size_t`) → implicit `int` in `send()` calls — Windows `send()` expects int for length, not `size_t`, and `/WX` promotes the warning to error. All the `(size_t)plen / (size_t)len` args in `socks5-client.c: 239, 335–336` and elsewhere (`output.c:378`, `connect4.c:99` etc) will hit the same issue.

saghul:
I know it wasn't your fault :-) Can you report it upstream? Can we remove /WX in the interim perhaps?

loss-and-quick:
> Can you report it upstream? Can we remove /WX in the interim perhaps?

Fixed in https://github.com/warmcat/libwebsockets/commit/122fd2481589a39fa0802df650ec90786a4e70e5

saghul:
Sweet!

saghul:
lws update: https://github.com/saghul/txiki.js/pull/993

saghul:
Merged the update. You can rebase now.

saghul (approved):

PR #972, opened 2026-06-11
sockets: add AbortSignal support for connect()

loss-and-quick:
# Summary

`tjs.connect()` and the underlying socket constructors (`TCPSocket`, `TLSSocket`, `PipeSocket`) now accept an optional `AbortSignal` to cancel in-flight connections.

```js
const ac = new AbortController();
setTimeout(() => ac.abort(new Error('timeout')), 3000);

const socket = await tjs.connect('tcp', 'example.com', 80, {
    signal: ac.signal,
});
```

# Changes

- **`BaseStreamSocket._connect(addr, signal)`** — Core logic:
  - Pre-aborted signals reject immediately, no connection attempt.
  - Abort during connection calls `silentClose(handle)` (uv_close) and rejects the promise.
  - Listener cleanup via `promise.finally()` — no leaks.

- **`PipeSocket(path, options)`** — Now accepts a second `options` argument (previously absent), forwards `options.signal` to `_connect()`.

- **`TCPSocket` / `TLSSocket`** — Forward `options.signal` to `_connect()`.

- **`tjs.connect()`** — Passes `options.signal` through to all constructors.

# Files changed

| File | Change |
|------|--------|
| `src/js/core/direct-sockets/utils.js` | `_connect(addr, signal)` — abort handling |
| `src/js/core/direct-sockets/tcp.js` | forward `options.signal` |
| `src/js/core/direct-sockets/tls.js` | forward `options.signal` |
| `src/js/core/direct-sockets/pipe.js` | `options` argument + forward `signal` |
| `src/js/core/sockets.js` | forward from `tjs.connect()` |
| `src/bundles/c/core/core.c` | rebuilt bundle (generated) |

# Testing

```js
// Pre-aborted — rejects immediately
const ac1 = new AbortController();
ac1.abort(new Error('cancelled'));
await tjs.connect('tcp', '127.0.0.1', 80, { signal: ac1.signal });
// → rejects with 'cancelled'

// Abort during connect — uv_close + reject
const ac2 = new AbortController();
setTimeout(() => ac2.abort(new Error('timeout')), 200);
await tjs.connect('tcp', '10.255.255.1', 9999, { signal: ac2.signal });
// → rejects with 'timeout' after ~200ms instead of blocking indefinitely


saghul:
Is this something you came up with or has it been added to the Direct Sockets spec?

loss-and-quick:
> Is this something you came up with or has it been added to the Direct Sockets spec?

This is an extension I implemented, same approach as Node.js `net.connect({ signal })` — cancels the connection via `uv_close`. The Direct Sockets spec doesn't define abort for the connect phase yet, but it's a natural extension since libuv already supports `uv_close` to cancel pending `uv_tcp_connect`.

saghul:
This needs some tests.

saghul (approved):
Can you rebase please?

PR #975, opened 2026-06-11
httpserver: restore query string in WS upgrade URI

loss-and-quick:
# Summary

`lws_http_get_uri_and_method()` returns only the URI path (e.g. `/ws`) and strips the query string (`?token=...`). The normal HTTP handler (`LWS_CALLBACK_HTTP`) already reconstructs the full URI via `WSI_TOKEN_HTTP_URI_ARGS`, but the WebSocket upgrade path (`LWS_CALLBACK_HTTP_CONFIRM_UPGRADE`) was passing only the bare path to the JavaScript `onRequest` handler.

This caused any JS code relying on `req.url` containing query parameters (e.g. `searchParams.get("token")`) to fail silently — `upgrade()` was never called and libwebsockets closed the connection with no response ("Empty reply from server").

## Root Cause

In `LWS_CALLBACK_HTTP_CONFIRM_UPGRADE` (~line 745), `args[2]` was set directly from `uri_ptr`:

```c
args[2] = JS_NewStringLen(ctx, uri_ptr, uri_len);
```

The same pattern in `LWS_CALLBACK_HTTP` (~line 798) correctly appends the query string:

```c
memcpy(req->url, uri_ptr, copy_len);
if (lws_hdr_copy(wsi, req->url + copy_len + 1, copy_query_len, WSI_TOKEN_HTTP_URI_ARGS) > 0) {
    req->url[copy_len] = '?';
}
```

The WS upgrade handler was simply missing this step.

## Fix

Applied the same URL reconstruction logic to the upgrade handler: copy `uri_ptr` into a local `full_url[2048]` buffer, then append `?` + the query string from `WSI_TOKEN_HTTP_URI_ARGS` when present. The constructed URL is passed to `args[2]` instead of the bare path.

## Testing

Verified on Android (arm64, Android 28) with a real WebSocket upgrade request:

```
> GET /ws?token=<uuid> HTTP/1.1
> Connection: Upgrade
> Upgrade: websocket
< HTTP/1.1 101 Switching Protocols
< Upgrade: WebSocket
< Connection: Upgrade
< Sec-WebSocket-Accept: s3pPLMBiTxaQ9kYGzzhZRbK+xOo=
```

Before the fix: `req.url` = `http://127.0.0.1:43125/ws` → token extraction returned `undefined` → upgrade rejected.

After the fix: `req.url` = `http://127.0.0.1:43125/ws?token=<uuid>` → token matches → HTTP 101, WebSocket data flows.

### DIFF
--- a/src/httpserver.c
+++ b/src/httpserver.c
@@ -740,11 +740,29 @@ static int tjs_http_callback(struct lws *wsi, enum lws_callback_reasons reason, 
 
             uint64_t upgrade_id = uctx->id;
 
+            /* Reconstruct full URL including query string (lws strips it from uri_ptr). */
+            int query_len = lws_hdr_total_length(wsi, WSI_TOKEN_HTTP_URI_ARGS);
+            int needs_sep = query_len > 0 ? 1 : 0;
+            char *full_url = tjs__malloc((size_t) uri_len + needs_sep + query_len + 1);
+            if (!full_url) {
+                return -1;
+            }
+            memcpy(full_url, uri_ptr, (size_t) uri_len);
+
+            if (query_len > 0 &&
+                lws_hdr_copy(wsi, full_url + uri_len + 1, query_len + 1, WSI_TOKEN_HTTP_URI_ARGS) > 0) {
+                full_url[uri_len] = '?';
+            } else {
+                full_url[uri_len] = '\0';
+            }
+
             /* Call JS onRequest synchronously with upgrade ID and WS marker. */
             JSValue args[7];
             args[0] = JS_NewInt64(ctx, (int64_t) upgrade_id);
             args[1] = JS_NewString(ctx, method);
-            args[2] = JS_NewStringLen(ctx, uri_ptr, uri_len);
+            args[2] = JS_NewStringLen(ctx, full_url, strlen(full_url));
+            tjs__free(full_url);
+
             args[3] = headers_arr;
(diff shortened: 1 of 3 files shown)


saghul:
Can you add a test for this please?

src/httpserver.c
```diff
@@ -740,11 +740,23 @@ static int tjs_http_callback(struct lws *wsi, enum lws_callback_reasons reason,
 
             uint64_t upgrade_id = uctx->id;
 
+            /* Reconstruct full URL including query string (lws strips it from uri_ptr). */
+            char full_url[2048];
```
saghul: Does lws define a max URL length constant? We could probably use that here.

src/httpserver.c
```diff
@@ -740,11 +740,29 @@ static int tjs_http_callback(struct lws *wsi, enum lws_callback_reasons reason,
 
             uint64_t upgrade_id = uctx->id;
 
+            /* Reconstruct full URL including query string (lws strips it from uri_ptr). */
+            int query_len = lws_hdr_total_length(wsi, WSI_TOKEN_HTTP_URI_ARGS);
+            int needs_sep = query_len > 0 ? 1 : 0;
+            char *full_url = malloc((size_t) uri_len + needs_sep + query_len + 1);
```
saghul: Use tjs__malloc / free please.

saghul (approved):

PR #976, opened 2026-06-11
fetch: add allowInsecure option for self-signed/expired TLS certs

loss-and-quick:
# Summary

Adds a new `allowInsecure` option to the `fetch()` API. When set to `true`, TLS certificate validation is relaxed — self-signed, expired, or otherwise untrusted certificates are accepted.

This is useful for development, testing, and internal networks where valid certificates are not available.

# Usage

```js
const res = await fetch('https://self-signed.local/api', {
  allowInsecure: true
});
```

# Implementation

- **`src/httpclient.c`** — Added `setAllowInsecure` native method that toggles the `LCCSCF_ALLOW_INSECURE` flag on the libwebsockets SSL connection (same flag used by the WebSocket client).
- **`src/js/polyfills/http-client.js`** — Added `#allowInsecure` private field and `setAllowInsecure()` method to the `HttpClient` JS wrapper. Propagates the flag to the native client both immediately and when the connection is opened.
- **`src/js/polyfills/fetch/fetch.js`** — Reads `init.allowInsecure` from the fetch options and passes it to the underlying `HttpClient`.

# Testing

```js
const res = await fetch('https://expired.badssl.com', {
  allowInsecure: true
});
console.log(res.status); // 200
```


src/httpclient.c
```diff
@@ -890,6 +890,20 @@ static JSValue tjs_httpclient_abort(JSContext *ctx, JSValue this_val, int argc,
     return JS_UNDEFINED;
 }
 
+static JSValue tjs_httpclient_set_allow_insecure(JSContext *ctx, JSValue this_val, int argc, JSValue *argv) {
+    TJSHttpClient *h = tjs_httpclient_get(ctx, this_val);
+    if (!h) {
+        return JS_EXCEPTION;
+    }
+
+    if (argc > 0 && JS_ToBool(ctx, argv[0])) {
```
saghul:
```suggestion
    if (JS_ToBool(ctx, argv[0])) {
```

saghul: Since you declared the function with 1 argument, it will be filled as undefined, which will evaluate to false anyway.

src/js/polyfills/fetch/fetch.js
```diff
@@ -98,6 +98,10 @@ export function fetch(input, init) {
 
         const client = new HttpClient();
 
+        if (init && init.allowInsecure) {
```
saghul:
```suggestion
        if (init?.allowInsecure) {
```

saghul (commented):
Left a couple of comments. Please add a test too.

bnoordhuis:
Drive-by comment: a more general solution is a `validateCertificate` callback (actual name TBD) that lets user code inspect and accept/reject the certificate. That way you can do things like certificate pinning.

(A secondary question is whether the callback should be sync or async.)

saghul:
> Drive-by comment: a more general solution is a `validateCertificate` callback (actual name TBD) that lets user code inspect and accept/reject the certificate. That way you can do things like certificate pinning.
> 
> (A secondary question is whether the callback should be sync or async.)

Thanks for chiming in Ben! Good idea indeed. @loss-and-quick can you take a look and see if lws has something like that we could tap into?

loss-and-quick:
Thanks, both comments applied (dropped the `argc > 0` guard, switched to `init?.allowInsecure`), and added a test (`test-fetch-allow-insecure.js`).

Re a `validateCertificate` callback — I looked into what lws exposes, and the short version is that it's **not cleanly available on the mbedtls backend that txiki builds against**:

- The natural hook is `LWS_CALLBACK_OPENSSL_PERFORM_SERVER_CERT_VERIFICATION` — but it's **OpenSSL-only**. It's wired up in `lib/tls/openssl/openssl-client.c` (passes the `X509_STORE_CTX` + `preverify_ok`, and the `wsi` is recoverable via `SSL_get_ex_data`, so per-connection routing to a JS callback would work there).
- On the **mbedtls** side there's no equivalent user hook: `lib/tls/mbedtls/mbedtls-client.c` only has the internal `lws_mbedtls_client_verify_callback` used for JIT-trust (AKID/SKID collection), which isn't surfaced to user code. Verification results are checked post-handshake in `lws_tls_client_confirm_peer_cert()` — too late to intercept the decision.
- `lws_tls_peer_cert_info()` (`LWS_TLS_CERT_INFO_*`, incl. `DER_RAW`) **does** work under mbedtls, but only **after** the handshake completes — usable for inspection/pinning-after-the-fact (connect, inspect, drop if unwanted), not for accepting/rejecting *during* the handshake.
- In all cases the verify path is **synchronous** — lws can't pause the handshake to await an async JS callback.

So a general `validateCertificate` would need either (a) switching txiki to OpenSSL, or (b) adding a new mbedtls hook upstream in lws. Both are well beyond this PR. I'd suggest landing `allowInsecure` as the pragmatic boolean now (it maps directly to `LCCSCF_ALLOW_INSECURE`), and tracking `validateCertificate` separately as a larger, backend-dependent feature. Happy to file an issue for it.

saghul:
mbedtls is at least partially wired through some openssl compatibility layer comming from some espressif code. Can you check if they also wired it there?

loss-and-quick:
> mbedtls is at least partially wired through some openssl compatibility layer comming from some espressif code. Can you check if they also wired it there?

Yes, I traced the verify callback chain through it, and the answer is: **they did not wire up** `LWS_CALLBACK_OPENSSL_PERFORM_SERVER_CERT_VERIFICATION` there either.

The root cause is signature mismatch. The Espressif wrapper's `SSL_set_verify` takes a callback with type:

```c
int (*)(SSL *ssl, mbedtls_x509_crt *x509)
```

not OpenSSL's:

```c
int (*)(int preverify_ok, X509_STORE_CTX *ctx)
```

The second parameter is a raw `mbedtls_x509_crt *`, not an `X509_STORE_CTX *`. In fact, in the wrapper `X509_STORE_CTX` is `typedef void X509_STORE_CTX`, and `X509_STORE_CTX_get_error()` is a stub that always returns `X509_V_ERR_UNSPECIFIED` — so even if you tried to pass one, it's useless.

The wrapper's mbedTLS verify bridge (`lws_mbedtls_f_vrfy` in `wrapper/platform/ssl_pm.c`) also ignores the `uint32_t *pflags` parameter and always returns 0 — so the callback is purely observational; it can't override the verification decision like OpenSSL's `X509_STORE_CTX_set_error()` can.

And lws's own mbedtls client code (`lib/tls/mbedtls/mbedtls-client.c`) doesn't use the OpenSSL callback path at all — it compiles separate code from `lib/tls/openssl/openssl-client.c` where callback 58 is defined. On the mbedtls side, verification is always handled post-handshake in `lws_tls_client_confirm_peer_cert()`.

So the situation stands as I described: a `validateCertificate` callback on mbedTLS would need a different design — probably registering in the mbedTLS verify hook (which fires per-cert during `SSL_connect`) and checking stashed results post-handshake, rather than using the OpenSSL callback 58 pattern.

tests/helpers/fetch-insecure-client.js
```diff
@@ -0,0 +1,14 @@
+// Helper for test-fetch-allow-insecure.js
```
saghul: Why create this helper and not run the fetch() in the test directly?

tests/helpers/tls-http-server.js
```diff
@@ -0,0 +1,21 @@
+// Minimal HTTPS server on a self-signed certificate for fetch tests.
```
saghul: Ditto, you can run the server directly in the test without needed an auxiliary process.

saghul (approved):

PR #980, opened 2026-06-12
fetch: honor init.timeout as a native request deadline

loss-and-quick:
## What

`fetch(url, { timeout })` (milliseconds) now arms the native `HttpClient.timeout` — the same plumbing `XMLHttpRequest.timeout` already uses. The lws scheduled-event timer is armed when the client connection is created and fires `LWS_CALLBACK_TIMER` from the service pass, settling the request with `TypeError: Network request timed out` if it hasn't completed by the deadline.

## Why not just AbortSignal.timeout()?

They work at different layers, and they compose:

- `AbortSignal` is driven by a JS timer, and the abort listener tears the wsi down from outside the lws callback (via `client.abort()`). That works while the connection is making progress, but a connection stuck in a native phase — mid-TLS-handshake, or behind a SOCKS5/HTTP proxy whose upstream is dead — could stay pending much longer than intended. We hit exactly this on Android: proxied requests through a dead upstream hung despite a signal.
- The lws timer fires from within the service loop itself, so it settles the wsi reliably regardless of which native phase the connection is stuck in.
- `xhr.timeout` already exposes this native deadline; `fetch` had no way to reach it.

`init.signal` continues to work exactly as before — `timeout` is just a deadline armed natively, and rejects with `TypeError` (network error) rather than `AbortError`, matching `xhr.ontimeout` semantics.

## Test

`tests/test-fetch-timeout.js` (modelled on `tests/test-xhr-timeout.js`): a `/delay/3` echo-server response with `timeout: 200` rejects with `TypeError`; a fast request with a generous timeout is unaffected.


saghul:
I don't want to extend fetch with options that are attainable from the outside, like using the signal.

Whatever state the native request is, it will be aborted. If there is anything not working there is should be fixed in the native layer.

PR #991, opened 2026-06-13
build: add size-optimization flags (STRIP, LTO, GC_SECTIONS, MinSizeRel)

lukasMega:
(discussion: https://github.com/saghul/txiki.js/discussions/954)
from original big PR https://github.com/saghul/txiki.js/pull/962 here is extracted only support for build optimization (CMake)  flags:

Add four CMake options that shrink the tjs binary without dropping any feature; they are orthogonal and can be combined:

- `BUILD_WITH_STRIP`: run `${CMAKE_STRIP}` on the binary as a post-build step (skipped when CMAKE_STRIP is unset, e.g. MSVC)
- `BUILD_WITH_LTO`: enable `CMAKE_INTERPROCEDURAL_OPTIMIZATION`, with a graceful warning when the toolchain lacks IPO support
- `BUILD_WITH_GC_SECTIONS`: compile with `-ffunction-sections`/`-fdata-sections` (`/Gy` on MSVC) and link with `--gc-sections` / `-dead_strip` / `/OPT:REF,/OPT:ICF`
- BUILDTYPE=MinSizeRel: documented; standard CMake build type (`-Os`)

Measured (macOS arm64): full Release 5.67 MB -> 3.71 MB with all four enabled, same feature set.

Document the flags in CLAUDE.md, README.md, and website/docs/building.m

---

my results (on macOS):

```console
% du -h build/tjs build-strip/tjs build-lto/tjs build-gc/tjs build-minsize/tjs build-opt/tjs
5.7M    build/tjs
5.2M    build-strip/tjs
5.3M    build-lto/tjs
5.1M    build-gc/tjs
4.6M    build-minsize/tjs
3.7M    build-opt/tjs     # All optimization flags, features kept: MinSizeRel + STRIP + LTO + GC_SECTIONS
```


website/docs/building.md
```diff
@@ -117,6 +117,42 @@ cmake -B build-slim -DCMAKE_BUILD_TYPE=Release -DBUILD_WITH_WASM=OFF -DBUILD_WIT
...
+
+Notes:
+
+- `BUILD_WITH_STRIP` runs the toolchain's `strip` as a post-build step. It is skipped where
+  `CMAKE_STRIP` is unset (e.g. MSVC).
+- `BUILD_WITH_LTO` falls back to a warning (not an error) if the toolchain cannot do
+  interprocedural optimization.
+- `BUILD_WITH_GC_SECTIONS` maps to `-Wl,--gc-sections` (GNU/lld), `-Wl,-dead_strip` (Apple), or
+  `/OPT:REF /OPT:ICF` (MSVC).
+- `BUILDTYPE=MinSizeRel` needs no extra flag; it is a standard CMake build type.
```
saghul: It uses `-Os` though on Unix, maybe worth mentioning.

website/docs/building.md
```diff
@@ -117,6 +117,42 @@ cmake -B build-slim -DCMAKE_BUILD_TYPE=Release -DBUILD_WITH_WASM=OFF -DBUILD_WIT
...
+
+Unix/macOS example combining all four:
```
saghul: I don't think this adds much, given the next example.

website/docs/building.md
```diff
@@ -117,6 +117,42 @@ cmake -B build-slim -DCMAKE_BUILD_TYPE=Release -DBUILD_WITH_WASM=OFF -DBUILD_WIT
...
+code is compiled and linked. They are independent of one another and can be combined.
+
+| CMake option                | Default | Effect                                              |
+|-----------------------------|---------|-----------------------------------------------------|
+| `BUILD_WITH_STRIP=ON`       | OFF     | Strip the symbol table from the binary after linking |
+| `BUILD_WITH_LTO=ON`         | OFF     | Enable link-time optimization (smaller/faster code, slower link) |
+| `BUILD_WITH_GC_SECTIONS=ON` | OFF     | Per-function/data sections plus linker dead-code stripping |
+| `BUILDTYPE=MinSizeRel`      | —       | Standard CMake build type that optimizes for size (`-Os`) |
+
+Notes:
```
saghul, line 132: Let's add a note here about the performance loss please.

README.md
```diff
@@ -34,6 +34,16 @@ BUILD_WITH_WASM=OFF make      # ~0.4 MB smaller — removes WebAssembly/WASI
 BUILD_WITH_SQLITE=OFF make    # ~1.5 MB smaller — removes the tjs:sqlite module
 ```
 
+The binary can also be shrunk without dropping any feature, via compiler/linker
```
saghul: Oh, I missed you added this part in the README before. Let's remove the whole section please, I want to keep the README quite minimal, and there is already a link to the website's building section right above.

saghul (changes requested):

README.md
lukasMega: ok, so I will remove whole section Slim build

saghul: Yes please.

website/docs/building.md
lukasMega: I will remove it.

website/docs/building.md
lukasMega: I added `::warning` section **Performance trade-off**  in my latest commit.
Preview: 
scr 2026-06-15 o 12 57 30

CMakeLists.txt
```diff
@@ -183,6 +205,23 @@ if(WIN32)
...
 endif()
 
+if(BUILD_WITH_STRIP AND CMAKE_STRIP)
+    add_custom_command(TARGET tjs-cli POST_BUILD
+        COMMAND ${CMAKE_STRIP} $<TARGET_FILE:tjs-cli>
+        COMMENT "Stripping tjs"
+    )
+endif()
+
+if(BUILD_WITH_GC_SECTIONS)
```
saghul, line 215: Why not merge this block with the one above, with the same gating condition?

CMakeLists.txt
```diff
@@ -94,6 +96,26 @@ if(BUILD_WITH_UBSAN)
...
+        set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
+    else()
+        message(WARNING "LTO requested but not supported: ${_ipo_output}")
+    endif()
+endif()
+
+xoption(BUILD_WITH_GC_SECTIONS "If ON, compile with per-function/data sections and link with dead-code stripping" OFF)
+if(BUILD_WITH_GC_SECTIONS)
+    if(MSVC)
+        add_compile_options(/Gy)
```
saghul, line 113: I noticed we don't set these per target, but global, is that intentional?

CMakeLists.txt
```diff
@@ -94,6 +96,26 @@ if(BUILD_WITH_UBSAN)
...
+    check_ipo_supported(RESULT _ipo_supported OUTPUT _ipo_output)
+    if(_ipo_supported)
+        message(STATUS "LTO enabled")
+        set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
+    else()
+        message(WARNING "LTO requested but not supported: ${_ipo_output}")
+    endif()
+endif()
+
+xoption(BUILD_WITH_GC_SECTIONS "If ON, compile with per-function/data sections and link with dead-code stripping" OFF)
```
saghul: Can you define the option together with the other ones?

CMakeLists.txt
lukasMega: yes, intentional. 
These compiler flags only pay off if every translation unit that ends up in the binary is compiled that way - including the deps. 
Putting them on `target_compile_options(tjs ...)` would section only our own objects and leave the bulk (the deps) un-sectioned, without the benefit.

CMakeLists.txt
lukasMega: If I understand it correctly, they can't share one block because of target ordering, not because of duplicated gating.

---

explanation from claude:

```
- The compile block (111–117) uses add_compile_options(...), which is directory-scoped. It sits before all  
  the add_subdirectory(deps/...) calls so the section flags propagate into the vendored deps (quickjs, libuv, 
  mbedtls, lws, wamr…). That's what makes --gc-sections actually worth anything — most of the strippable dead 
  code lives in those deps.                                                                                   
  - The link block (215–223) uses target_link_options(tjs-cli ...), and tjs-cli doesn't exist until line 195. 
                                                                                                              
  If we moved the compile block down next to the link block at 215, it would land after the deps              
  subdirectories are added, so the deps would no longer be compiled with -ffunction-sections/-fdata-sections  
  and the linker would have far less to strip. So the split is intentional: compile flags early (to reach     
  deps), link flag on the final executable. 
```

website/docs/building.md
saghul: Sounds good, thanks!

saghul:
can you rebase please?

saghul:
Something went weird with the merge, I see some lws related commits.

lukasMega:
oh, I used to do merges most of the time, but here I tried rebase. I will try to clean up, I apologize.
 
//update: weird, in Idea it looks ok:

<details><summary>click to see</summary>

scr 2026-06-15 o 17 34 32

</details>

lukasMega:
I will try to recreate it into clean new branch
//update: https://github.com/saghul/txiki.js/pull/996

PR #992, opened 2026-06-13
build: add BUILD_WITH_TLS option to make TLS optional

lukasMega:
(discussion: https://github.com/saghul/txiki.js/discussions/954)
from original big PR https://github.com/saghul/txiki.js/pull/962 here is extracted only support for removing TLS in build using option `BUILD_WITH_TLS`:

Gate the TLS protocol layer behind a new CMake option `BUILD_WITH_TLS` (default ON), mirroring the existing `BUILD_WITH_WASM` / `BUILD_WITH_SQLITE` splits and their `TJS_HAVE_*` "define-when-ON" convention.

When OFF, the build drops `mod_tls.c`, the `mbedtls/mbedx509` libraries, and the libwebsockets SSL layer (LWS_WITH_SSL / LWS_WITH_MBEDTLS). libmbedcrypto is always linked, so the Web Crypto API (crypto.subtle) is unaffected. HTTPS, WSS, and TLSSocket/TLSServerSocket throw "… not supported in this build" on a NO_TLS binary; plain HTTP/WS and TCP/UDP keep working.

The active feature set is exposed via `tjs.engine.features.tls`, and the test runner skips TLS-dependent tests (test-tls-*, test-dispose-tls, test-fetch-allow-insecure) via `tests/feature-skip.json`.

Verified on macOS arm64: `TLS=ON 255/255 tests pass;` `TLS=OFF 246 pass / 9 skip / 0 fail`, `mbedtls_ssl_*` symbols absent while `mbedtls_aes_*` remain, binary 5.7 MB -> 5.2 MB.

---

my results (on macOS):

```console
% du -h  build/tjs build-notls/tjs 
5.7M    build/tjs
5.2M    build-notls/tjs
```


saghul:
I'm a bit on the fence on this one. It's the most invasive of the bunch, and the gain is not that huge.

I want to hold off until I see it more clearly.

lukasMega:
Ok, no problem. I just wanted to show it. Feel free to play with it, add changes or close it.

PR #998, opened 2026-06-16
docs: add instructions for mise

lukasMega:
# add section how to use with `mise` ( https://mise.jdx.dev/ )

Screen Shot 2026-06-16 at 22 59 53-fullpage

### DIFF
--- a/website/sidebars.ts
+++ b/website/sidebars.ts
@@ -41,7 +41,6 @@ const sidebars: SidebarsConfig = {
       label: 'Guides',
       collapsed: false,
       items: [
-        'guides/mise',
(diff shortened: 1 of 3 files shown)


website/sidebars.ts
```diff
@@ -41,6 +41,7 @@ const sidebars: SidebarsConfig = {
       label: 'Guides',
       collapsed: false,
       items: [
+        'guides/mise',
```
lukasMega: I'm not sure in which position to place it

saghul: Let's put it at the end please

PR #999, opened 2026-06-16
Typescript integration

KaruroChori:
It has been a while!
I finally implemented something which justified my initial attempt at integrating wasm in txiki (I am very happy someone else took over, I would have not had the time to be honest, this year has been a lot)

Basically, I wanted to run fast type elision on typescript files, so that I no longer needed an explicit transpilation step.

To be totally honest, I don't expect this to be merged for three reasons:

1) The code is mostly LLM generated and while I reviewed part of it, I don't really want to commit more effort if there is no interest in this feature. While it does not look too bad and it works just fine... I would not qualify it as more than a proof of concept.
2) It bundles a very narrow rust dependency, as all typescript transpilers these days are in rust or go. This is only needed to generate wasm bundles that can be shared in different ways. In theory it would be possible to push that part outside the main repo.
3) It introduces latency, mostly based on the mode WAMR has been compiled. I added a flag to support Fast JIT mode as it has the least amount of dependencies.

| File size    | Lines | Baseline (JS) | Interpreter | Fast JIT |
|-------------|------|--------------|-------------|----------|
| 0.1 KB      | 3    | 0.003 ms     | 0.16 ms     | 0.04 ms  |
| 0.5 KB      | 10   | 0.020 ms     | 0.60 ms     | 0.10 ms  |
| 2.7 KB      | 50   | 0.090 ms     | 2.80 ms     | 0.42 ms  |
| 11.4 KB     | 200  | 0.380 ms     | 11.0 ms     | 1.50 ms  |

Performance should be better for LLVM JIT and AOT, but they both require extra dependencies I did not want to test.  
These are not the most scientific tests out there, you can try the new script in benchmarks yourself.

The documentation has been aligned to explain a bit what is going on and how to use the typescript integration but basically:

- It can be disabled fully.
- The WASM module gets embedded in the binary itself.
- The WASM module is searched for in a HOME subpath. If not found an attempt is done to download it linked to the same release number from git.

Being able to use oxc or any other alternative is that it opens the path to handle jsx and tsx files as well. Or to minify and bundle a codebase without extra tools.

## How to use

Create a typescript file and just run it as you would for a normal js file.


KaruroChori:
[oxc_transpiler.wasm.zip](https://github.com/user-attachments/files/29022955/oxc_transpiler.wasm.zip)

If you trust strangers with binaries (well it is wasm, but wasi breaks the sandbox quite a bit) and you don't want rust on your system, you can copy this in ` ~/.tjs/typescript/{version}/oxc_transpiler.wasm ` with version the current one in master, which should be 26.5.0 but I am not 100% sure. 
So far in my test I used it bundled and compiled within the txiki binary, not as a separate file, that path has not been fully tested yet.

All artifacts in here
https://github.com/KaruroChori/txiki.js/actions/runs/27652157373

and the related "fake" release https://github.com/KaruroChori/txiki.js/releases/tag/v26.5.0

Experimental tsx support on a separate branch https://github.com/KaruroChori/txiki.js/tree/feature/tsx-integration

saghul:
Hey there! This is fantastic, I love it!

I'll have a full review tonight / tomorrow.

.github/workflows/release.yml
```diff
@@ -89,8 +89,25 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          submodules: recursive
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          target: wasm32-wasip1
+      - name: Build WASM transpiler
+        run: make oxc
+      - name: Upload WASM transpiler artifact
```
saghul: I think to keep things simple I can create a new repo and have this in the releases. This way the oxc build is independent of the txiki.js release, similar to what we do with esbuild.

src/js/run-main/index.js
```diff
@@ -76,7 +76,7 @@ Options:
 
 Subcommands:
   run
-        Run a JavaScript program
+        Run a JavaScript/TypeScript program
```
saghul:
```suggestion
        Run a JavaScript or TypeScript program
```

src/js/run-main/index.js
```diff
@@ -303,6 +312,8 @@ if (!isBundled) {
                     tjs.setImportMap(mapObj, path.dirname(resolvedMapPath));
                 }
 
+                // .ts, .tsx, .mts, .cts files are transpiled by the
```
saghul: what are .mts and .cts ? If they are Node-isms to support ESM vs CJS module loading, I'd say we skip those. TS means ESM really.

src/js/run-main/index.js
```diff
@@ -363,7 +374,16 @@ if (!isBundled) {
             }
 
             const infilePath = path.parse(infile);
-            const data = await tjs.readFile(infile);
+            let data = await tjs.readFile(infile);
+            const inExt = path.extname(infile).toLowerCase();
+
+            if (inExt === '.ts' || inExt === '.tsx' || inExt === '.mts' || inExt === '.cts') {
+                const { transpile } = await import('tjs:typescript');
```
saghul, line 381: Another option here would be to use esbuild even when compiling. Maybe something to think about for the future.

src/js/stdlib/typescript.js
```diff
@@ -0,0 +1,153 @@
+/* global tjs */
+import core from 'tjs:internal/core';
+
+let transpilerInstance = null;
+let transpilerModule = null;
+
+async function loadEmbedded() {
```
saghul, line 9: Not a fan of this, let's drop it an always use the external version instead please.

src/js/stdlib/typescript.js
```diff
@@ -0,0 +1,153 @@
...
+
+    if (bytes) {
+        return compileAndInstantiate(bytes);
+    }
+
+    return false;
+}
+
+async function loadFromCache() {
+    const tjsHome = (tjs.env && tjs.env.TJS_HOME) || tjs.homeDir + '/.tjs';
```
saghul, line 20: I think I added a helper to give you the tjs sone. Also you want to use the path module here to make sure it works properly on Windows.

src/js/stdlib/typescript.js
```diff
@@ -0,0 +1,153 @@
...
+    await tjs.makeDir(cacheDir, { recursive: true });
+    await tjs.writeFile(wasmPath, wasmBytes);
+
+    return await compileAndInstantiate(wasmBytes);
+}
+
+async function compileAndInstantiate(wasmBytes) {
+    transpilerModule = new WebAssembly.Module(wasmBytes);
+
+    const wasiImport = {
```
saghul: Why do we need to do this? I'm not even sure it's fully functional since it's not using `tjs:wasi`.

src/js/stdlib/typescript.js
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    if (await loadFromCache()) {
+        return true;
+    }
+
+    return await downloadAndCache();
+}
+
+await ensureTranspiler();
```
saghul, line 93: Wouldn't it make sense to do this lazily the first time `transpile` is used?

src/builtins.c
```diff
@@ -38,6 +38,9 @@
 #endif
 #include "bundles/c/stdlib/utils.c"
 #include "bundles/c/stdlib/uuid.c"
+#ifdef TJS_HAVE_TYPESCRIPT
```
saghul, line 42: I wouldn't gate this since it's really small (when the wasm is dynamically downloaded, that is).

src/modules.c
```diff
@@ -184,25 +186,93 @@ JSModuleDef *tjs_module_loader(JSContext *ctx, const char *module_name, void *op
             goto after_load;
         }
         use_realpath = false;
+        effective_name = module_name;
     } else {
         r = tjs__load_file(ctx, &dbuf, module_name);
+        if (r != 0) {
+            /* Try extension resolution for path-like imports with no extension. */
```
saghul, line 193: Here we are trying to solve the `import './foo';` case, where `foo` is really `foo.ts` that will eventually be transpiled to `foo.js` right? I thought that was frowned upon now?

src/modules.c
```diff
@@ -184,25 +186,93 @@ JSModuleDef *tjs_module_loader(JSContext *ctx, const char *module_name, void *op
...
+            JSValue args[2] = {
+                JS_NewString(ctx, effective_name),
+                JS_NewStringLen(ctx, (char *)dbuf.buf, dbuf.size - 1)
+            };
+            JSValue result = JS_Call(ctx, qrt->builtins.typescript_transpiler,
+                                      JS_UNDEFINED, 2, args);
+            JS_FreeValue(ctx, args[0]);
+            JS_FreeValue(ctx, args[1]);
+            if (JS_IsException(result)) {
+                if (type != JS_IMPORT_TYPE_BYTES) {
```
saghul, line 245: This is tautological, because if above says `if (type == JS_IMPORT_TYPE_JS`

src/modules.c
```diff
@@ -184,25 +186,93 @@ JSModuleDef *tjs_module_loader(JSContext *ctx, const char *module_name, void *op
...
+            size_t tlen;
+            tsrc = JS_ToCStringLen(ctx, &tlen, result);
+            if (tsrc) {
+                tbuf_free(&dbuf);
+                tbuf_init(ctx, &dbuf);
+                tbuf_put(&dbuf, (const uint8_t *)tsrc, tlen);
+                tbuf_putc(&dbuf, '\0');
+                JS_FreeCString(ctx, tsrc);
+            }
+            JS_FreeValue(ctx, result);
```
saghul, line 260: We should probably return an error here, no?

src/vm.c
```diff
@@ -231,6 +231,21 @@ static JSValue tjs__set_import_map_resolver(JSContext *ctx, JSValue this_val, in
     return JS_UNDEFINED;
 }
 
+static JSValue tjs__set_typescript_transpiler(JSContext *ctx, JSValue this_val, int argc, JSValue *argv) {
+    TJSRuntime *qrt = TJS_GetRuntime(ctx);
+    CHECK_NOT_NULL(qrt);
+
+    JS_FreeValue(ctx, qrt->builtins.typescript_transpiler);
+
+    if (argc > 0 && JS_IsFunction(ctx, argv[0])) {
```
saghul, line 240: You don't need to check if `argc > 0` if you declared the function length to be 1, the engine fills it with `undefined` already.

src/vm.c
```diff
@@ -934,6 +957,14 @@ JSValue TJS_EvalScript(JSContext *ctx, const char *filename) {
     return ret;
 }
 
+static bool tjs__has_ts_suffix(const char *str) {
```
saghul, line 960: This could probably go elsewhere and be reused in the module loader code.

src/vm.c
```diff
@@ -953,7 +984,36 @@ JSValue TJS_EvalModule(JSContext *ctx, const char *filename, bool is_main) {
...
+        if (JS_IsFunction(ctx, qrt->builtins.typescript_transpiler)) {
+            JSValue args[2] = {
+                JS_NewString(ctx, filename),
+                JS_NewStringLen(ctx, (char *)dbuf.buf, dbuf_size)
+            };
+            JSValue result = JS_Call(ctx, qrt->builtins.typescript_transpiler,
+                                      JS_UNDEFINED, 2, args);
+            JS_FreeValue(ctx, args[0]);
+            JS_FreeValue(ctx, args[1]);
+            if (!JS_IsException(result)) {
```
saghul, line 992: We should let the exception bubble up here, in case there is one.

website/docs/guides/code-bundling.md
```diff
@@ -28,7 +28,9 @@ tjs bundle my-app/index.js
 
 ### TypeScript support
 
-esbuild natively strips TypeScript types during bundling, so `.ts` and `.tsx` files work as input without any extra configuration. Note that esbuild does **not** perform type checking. You should use `tsc --noEmit` separately if you need that.
+txiki.js runs `.ts` and `.tsx` files natively — no bundling required for development. The built-in WASM transpiler strips types at module load time.
+
+For production deployment, `tjs bundle` also works with TypeScript input:
```
saghul, line 35: Let's keep the esbuild mention here when `tjs bundle` is used.

website/docs/building.md
```diff
@@ -92,17 +92,31 @@ The executable will be at `build\Release\tjs.exe`.
...
+| `BUILD_WITH_WAMR_FAST_JIT=ON`      | OFF     | JIT-compile WASM for 4-8× faster transpilation | —              |
 
 When WebAssembly is disabled, the `WebAssembly` global is not installed and the `tjs:wasi`
 module is not available. When SQLite is disabled, the `tjs:sqlite` module is not available
 and `localStorage` falls back to a non-persistent, in-memory store (`sessionStorage` is
 unaffected). The active set of feature flags is exposed to JS via `tjs.engine.features`
 (e.g. `tjs.engine.features.wasm`, `tjs.engine.features.sqlite`).
 
+**TypeScript**: When enabled (default), the WASM transpiler is downloaded on first use and
+cached in `~/.tjs/typescript/<version>/`.  `BUILD_WITH_TYPESCRIPT=EMBED` compiles it
```
saghul, line 110: Let's drop the embedding please.

website/docs/building.md
```diff
@@ -92,17 +92,31 @@ The executable will be at `build\Release\tjs.exe`.
...
+directly into the binary at the cost of ~1.3 MB.  Set `BUILD_WITH_TYPESCRIPT=OFF` to
+disable TypeScript support entirely.
+
+**WAMR Fast JIT**: The WAMR WebAssembly runtime ships with a lightweight JIT compiler
```
saghul, line 114: Can we perhaps have a toggle to turn this on?

website/docs/building.md
```diff
@@ -92,17 +92,31 @@ The executable will be at `build\Release\tjs.exe`.
...
+(no LLVM needed — it uses [asmjit](https://asmjit.com/)).  Enabling it compiles hot WASM
```
saghul, line 115: It seems to support more architectures? "It supports X86, X86_64, and AArch64 architectures,"

website/docs/building.md
```diff
@@ -114,9 +128,22 @@ Direct CMake example (the flags can be combined):
 
 ```bash
 cmake -B build-slim -DCMAKE_BUILD_TYPE=Release -DBUILD_WITH_WASM=OFF -DBUILD_WITH_SQLITE=OFF
-cmake --build build-slim
```
saghul, line 117: Leave this in please.

website/docs/building.md
```diff
@@ -114,9 +128,22 @@ Direct CMake example (the flags can be combined):
...
 ```
 
+### WASM transpiler build
```
saghul: Let's move this to another repo, I'll create one.

website/docs/typescript.md
```diff
@@ -5,18 +5,105 @@ title: TypeScript
...
+## Import Resolution
+
+Import statements with explicit `.ts` extensions work directly:
+
+```ts
+import { foo } from './bar.ts';
+```
+
+For extensionless imports, txiki.js automatically tries resolving with these extensions in order:
+`.ts`, `.tsx`, `.mts`, `.cts`, `.js`, `.mjs`, `.cjs`.
```
saghul: I think `.ts`, `.tsx`,`.js` is enough. We never supported the others really.

saghul (changes requested):

saghul:
@KaruroChori Let's put the oxc transpiler in https://github.com/saghul/txikijs-ts-transpiler so the builds are independent.

saghul:
> I finally implemented something which justified my initial attempt at integrating wasm in txiki (I am very happy someone else took over, I would have not had the time to be honest, this year has been a lot)

Thanks for getting that going! I leveraged LLMs to get it over the finish line, and so far so good!



> To be totally honest, I don't expect this to be merged for three reasons:
> 

Why not? I think it's great! Some changes are necessary, sure, but I find the principle sound and would like to move forward with it.

>     1. The code is mostly LLM generated and while I reviewed part of it, I don't really want to commit more effort if there is no interest in this feature. While it does not look too bad and it works just fine... I would not qualify it as more than a proof of concept.
> 

I do think this is worth pursuing. Let me know if you don't have the time / interest and I can take over if you prefer.

>     2. It bundles a very narrow rust dependency, as all typescript transpilers these days are in rust or go. This is only needed to generate wasm bundles that can be shared in different ways. In theory it would be possible to push that part outside the main repo.
> 

Yeah my thoughts exactly. I created https://github.com/saghul/txikijs-ts-transpiler for that. You can start by PR-ing the transpiler build there, so we have it available for this PR to use.

>     3. It introduces latency, mostly based on the mode WAMR has been compiled. I added a flag to support Fast JIT mode as it has the least amount of dependencies.

I think that is very much acceptable, since I consider TS support to be useful for rapid prototyping (we should put that in the docs) but for a real app one should bundle it.

I missed the existence of Fast JIT, I think it's great to have the option to turn it on. In fact I'd probably make it default ON. Those who want small builds will likely disable WASM completely anyway.

KaruroChori:
@saghul would you consider adding an organisation associated with txiki to host all of its "official" content?
Mostly just as a namespace to keep things clean, but I have no strong feelings about this, either way it works for me :)

> Let me know if you don't have the time / interest and I can take over if you prefer.

I have a pretty heavy contract covering the next 5 weeks, but I *should* have enough time to push this forward and after that it should be no problem. 

> I missed the existence of Fast JIT, I think it's great to have the option to turn it on. In fact I'd probably make it default ON. Those who want small builds will likely disable WASM completely anyway.

I would suggest against having it on by default.
That feature is self-standing, but my understanding is that it has limited platform and OS support, being based on asmjit.
https://asmjit.com/
https://bytecodealliance.github.io/wamr.dev/blog/introduction-to-wamr-running-modes/
This blogpost for this feature is quite old, so compatibility might be better now, but 32 bit arm and riscv are not supported for sure. Turning it on by default might break builds for too many people.

website/docs/building.md
KaruroChori: I am not 100% sure supported platform for asmjit are always reflected in support by WAMR. I would have to check the downstream documentation for that. Also, there is the concern of supported OS, the material from WAMR I found only reports unix-based systems.

website/docs/building.md
KaruroChori: Are you sure? Not enabling this mode prevents shipping a single binary, and forces the user to have an internet connection first time they request the transpiler.
I realise it might be a marginal usage, but I would be forced to patch this in my distribution if left out.

src/js/run-main/index.js
KaruroChori: They are extensions which were common in the transition period to distinguish between common js (require) and module versions (import).
Technically, part of the role of a transpiler could be to reduce legacy require into imports or the other way around, hoping not to break everything. That is something bun does if I recall correctly.

In the context of txiki js==mjs and ts == mts I suppose.
But yes, we can drop them.

src/js/stdlib/typescript.js
KaruroChori: It is not meant to be functional. At times CLI tools have expectations about posix functions being available when building, even if not actively used for the functions which will be called. This it a basic polyfill used to silence such cases.

However, I just checked and it seems there is no reason to use it here; the way I integrated oxc does not require the wasi polyfill it seems.

KaruroChori:
I had a quick look and added some comments/extra information for various points which have been raised in review.

I think the first "priority" for this PR is to split off the oxc stuff into a separate repo and get releases working there, so that proper CI checks can resume here.
The rest of the changes are not "architectural" so they should be addressable without too many problems.

saghul:
> @saghul would you consider adding an organisation associated with txiki to host all of its "official" content?
> Mostly just as a namespace to keep things clean, but I have no strong feelings about this, either way it works for me :)

I did consider that, let me sleep on it a bit :-) For now there is that oxc repo, but I can transfer it later too.

houd1ni:
Why using another compiler at all ? we are already dependent on esbuild with no troubles. 🤷 
really looks like huge complexity boost gaining weight of the runtime. also wasm most certainly would not outperform esbuild executable.

@KaruroChori hi! were you aware of `tjs bundle` feature ?
it could run in the background in watch (dev) mode and compile-an-run in production with all the features and flags, including sourcemaps.

PS as my experience with Oxc with vite, it's a bit of mess for now. I mean feature support and common maturity.

saghul:
The use of oxc is for type stripping, not bundling, so while related it's not the same.

houd1ni:
> The use of oxc is for type stripping, not bundling, so while related it's not the same.

yep, and esbuild already covers that.

btw even fat nodejs does not carry a compiler for type stripping afaik. although its' abilities there are pretty limited to 'just' stripping with no support for runtime constructions that typescript makes, hence it's almost unusable with that stripping.

PS 4 me personally the pain is all about weight and especially complexity:
- looks like it was a lot of work on top of llm help (my bad, that haven't seen this before tho)
- it takes one more similar thing to support.

@saghul are you keeping in mind to use that mechanism of merging wasm binaries for other stuff ? maybe this one could be an interesting platform to start with. And what are weak points in using esbuild here ? Perhaps I'm not seeing an elephant. thx!

KaruroChori:
@houd1ni truth is that I was initially unaware of `tjs bundle`, as it was added only "recently"; for several reasons, I am using a custom fork of txiki for my projects, which is a bit in the past; so I found out about the esbuild integration only while writing code for this PR.

It is true esbuild allows for transpilation.
Still, l thought it was a good idea to proceed this way for the following reasons:

- `esbuild` distributed that way is a binary which can only be interacted with its CLI interface. By comparison I think the WASM integration I proposed is more elegant, as you "exchange" buffers of memory between txiki and the wasm runtime without having to move data through all those layers.
- The size of the binary being downloaded is quite hefty, even more so once unzipped. And it must be loaded in memory each time we want to run it which takes while; this might be a bit too much for some of the embedded systems I work with. On a linux device with 64MB of RAM, wasting 10 just for that is not ideal. But I realise, this is mostly a "me problem".
- The only mechanism which allows to use esbuild expects internet connectivity available on the target device, and a centralized package manager to provide it. 
- I am not sure what is the full extent of architectures and operating systems for which binaries of esbuild are made available.

I understand that a lot of these concerns might not be "typical" :)

> saghul are you keeping in mind to use that mechanism of merging wasm binaries for other stuff ? 

My understanding is that he does not like that much, but it is desirable in my use-case; it is likely I will drop support for it in this PR, and keep it within my fork.

saghul:
That's pretty much right.

I like the idea of WASM modules we could download. Technically esbuild can also be built with WASM, and might be interesting for the future.

saghul:
Another potential candidate: https://github.com/theMackabu/skim

KaruroChori:
> Another potential candidate: https://github.com/theMackabu/skim

It surely has the advantage of being small and more trivially embeddable, and if implemented properly it is probably faster. 
My concern is its recency and the somewhat limited test-suite. As no one else has adopted it yet, it would probably need some benchmarking and stronger evaluation against real world codebases I suppose.

PR #1015, opened 2026-06-29
ffi: fix size_t formatting

brainrom:
Fix the format specifier in `src/mod_ffi.c:510`. The value of type `size_t` was formatted with `%lu`, which is incorrect on platforms where `size_t` have different underlying type . On 32-bit architectures (e.g., i686, armv7, etc.), `size_t` is `unsigned int`, while `%lu` excepts `long unsigned int`. Even if they same byte size, they are distinct types in C, so `%lu` triggers a compiler error. Replaced with the correct `%zu` specifier, as per C99 standard for `size_t`.


saghul (approved):

saghul:
Thanks!

PR #1028, opened 2026-07-11
process: fix subprocess stdin pipe jamming after the first write

### LINKED ISSUE #1027: tjs.spawn(..., { stdin: 'pipe' }): only the first stdin write is delivered, write promises never resolve
Note: this bug was found, root-caused, and written up by an AI agent (Claude Code / Claude Fable 5). I double checked the problem, and read through this multiple times to double check the problem. Waiting a few days before publishing as to not add to AI slop. Please ignore if you want.

# Symptom

With a subprocess spawned with stdin: 'pipe', only the first write to proc.stdin reaches the child. The promise returned by writer.write() never resolves, so the WritableStream's internal queue jams and every subsequent write is queued forever.

# Reproduction

```js
const proc = tjs.spawn(['cat'], { stdin: 'pipe', stdout: 'pipe' });
const writer = proc.stdin.getWriter();
const enc = new TextEncoder();

await writer.write(enc.encode('one\n')); // delivered, but this await never returns
await writer.write(enc.encode('two\n')); // never even starts
```

Reading the child's stdout shows it received `"one\n"` only; neither write promise ever settles.

# Root cause 

`ProcessWritableStream`'s sink in `src/js/core/process.js` decides whether `handle.write` (chunk) completed synchronously with `typeof result !== 'number'`. That check matches the UDP contract — `tjs_udp_send returns` the byte count (`JS_NewInt64`) inline and undefined when an async send was queued — and appears to have been carried over from there in `ccbb79c` ("process: consolidate code for subprocess stdio handles").

But the native method it actually calls, `tjs_stream_write` (`src/mod_streams.c`), returns booleans: `JS_TRUE` when `uv_try_write` delivered the whole chunk inline, `JS_FALSE` when an async `uv_write` was queued (whose completion later fires onwrite).

So on the common fast path `write()` returns true, the sink misreads that as "async write pending", pushes a resolver onto the onwrite queue, and awaits a promise nothing will ever resolve — no async write exists, so onwrite never fires. The first chunk was delivered (the try-write succeeded), which is what makes it look like "only the first write works".

For comparison, `src/js/core/direct-sockets/utils.js` consumes the same native API with the correct boolean check: `if (this.#handle.write(buf)) { return Promise.resolve(); }`

# Fix

One line in the sink: `if (result === false) {`. Verified locally: with the fix, all write promises resolve and the child receives every chunk. PR incoming.

# Environment

txiki.js master (reproduced at `fa9340a`), macOS arm64; the bug is platform-independent (it triggers whenever `uv_try_write` accepts the whole chunk, which is the normal case for small writes to a pipe).

tarwin:
Note: this fix was authored by an AI agent (Claude Code / Claude Fable 5). I went over the changes multiple times to confirm. Understand if you want to ignore as AI slop. I wasn't sure if `core.c` should be included or not but left as is.

# Fixes #1027

# Problem

With `tjs.spawn(..., { stdin: 'pipe' })`, only the first write to the child's stdin was delivered; its promise never resolved and every subsequent write queued forever.

# Cause

The `ProcessWritableStream` sink checked `typeof result !== 'number'` to decide whether handle.write(chunk) completed synchronously. That matches the UDP contract and was carried over in `ccbb79c` — but `tjs_stream_write` returns booleans: true when `uv_try_write` delivered the whole chunk inline, `false` when an async `uv_write` was queued whose completion fires onwrite. On the common fast path `write()` returned true, the sink misread that as "async write pending" and awaited an onwrite completion that never came. The first chunk was still delivered, which made the jam look like "only the first write works".

# Fix

Check for the boolean the native method actually returns `(result === false)`, matching how `direct-sockets/utils.js` already consumes the same API.

# Tests 
 
`tests/test-spawn-stdin-pipe-multiple-writes.js` writes three chunks to a child's piped stdin, awaits each write promise, and asserts the child echoed all of them back (hangs on the second write without the fix).

`tests/helpers/stdin-cat.js` is a portable cat (tjs.stdin.pipeTo(tjs.stdout)) run via tjs.exePath, so the test also works on Windows. Full test suite passes locally (macOS arm64). Includes the rebuilt src/bundles/c/core/core.c.


src/js/core/process.js
```diff
@@ -118,9 +118,12 @@ class ProcessWritableStream extends WritableStream {
         super({
             async write(chunk, controller) {
                 try {
+                    // write() returns true when the chunk was fully written
+                    // inline (uv_try_write) and false when an async write was
+                    // queued, in which case onwrite fires on completion.
                     const result = handle.write(chunk);
 
-                    if (typeof result !== 'number') {
+                    if (result === false) {
```
saghul: Can you use !result instead?

saghul (commented):
Thanks for the fix! Left a small comment.

src/js/core/process.js
tarwin: Could do that. Although I do feel that being explicit about "false" is better here given that we would not expect just generally "falsey" results. But I'm very much not wedded to this decision.

saghul: I see your point but I feel like if we change the internal API breaking this we'd need to go through all parts of the code which use it anyway.

Can you change it please? :-)

tarwin: Sorry for the delay. I've updated the pull request.

saghul (approved):

PR #1076, opened 2026-08-24
builtins: throw on unknown tjs: module instead of failing silently

lukasMega:
### Problem

Importing a `tjs:` module that doesn't exist doesn't report an error.

```js
import 'tjs:doesnotexist';
console.log('reached');
```

This prints nothing and exits with status **0**. The script stops at the import, but nothing says why, and CI sees a success.

The dynamic form is no better:

```js
let caught;
try {
    await import('tjs:doesnotexist');
} catch (e) {
    caught = e;
}
```

`caught` ends up holding QuickJS's internal `JS_UNINITIALIZED` sentinel, so merely touching it throws `ReferenceError: caught is not initialized` — an error about the wrong thing entirely.

The cause is in `tjs__load_builtin()` (`src/builtins.c`): when no builtin matches it returns `NULL` without setting an exception, and the QuickJS module-loader contract requires a pending exception whenever a loader returns `NULL`. Both call sites in `src/modules.c` return the result directly, so the bare `NULL` propagates straight out of `tjs_module_loader()`.

### Second bug in the same function

The name lookup used `strncmp(p->name, name, strlen(p->name))`, which compares only as many characters as the *registered* name has — so it matches on a prefix. Any specifier starting with a real module name silently resolved to that module:

```js
import path from 'tjs:pathological';  // actually gives you tjs:path
```

### Fix

Two lines:

- `strncmp(...)` → `strcmp(...)`, so a module name has to match exactly.
- Throw `ReferenceError: could not load '<specifier>'` before returning `NULL`.

The static import above now exits non-zero with a `ReferenceError` naming the specifier, and the dynamic import rejects with a catchable `ReferenceError`.

### Tests

Three files. The dynamic case is tested in-process with `try`/`catch`. The static case has to run out of process — a failing static import aborts the whole script, so it can't be caught — and asserts the exit status, the error type, the specifier in the message, and that execution never reaches the line after the import.

Full suite passes (379 OK / 0 FAIL). Both changes are mutation-checked: reverting either one makes the new tests fail.

### DIFF
--- a/src/builtins.c
+++ b/src/builtins.c
@@ -77,13 +77,14 @@ JSModuleDef *tjs__load_builtin(JSContext *ctx, const char *name) {
     tjs_builtin_t *item = NULL;
 
     for (tjs_builtin_t *p = builtins; p->name != NULL; ++p) {
-        if (strncmp(p->name, name, strlen(p->name)) == 0) {
+        if (strcmp(p->name, name) == 0) {
             item = p;
             break;
         }
     }
 
     if (item == NULL) {
+        JS_ThrowReferenceError(ctx, "could not load '%s'", name);
         return NULL;
     }
 
--- a/tests/helpers/import-unknown-builtin.js
+++ b/tests/helpers/import-unknown-builtin.js
@@ -1,0 +1,3 @@
+import 'tjs:doesnotexist';
+
+console.log('reached');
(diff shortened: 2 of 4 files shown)


saghul (approved):

saghul:
Nice!

